PR #44, opened 2017-10-26
V4 sample

dosumis:
- [x] Fix merge conflicts
- [x] Reconcile work test scripts and structure of tests  - @dosumis
Decided to keep Simon's test script as-is for now.  Should not be affected by changes I've made.  Will simplify later.
- [x] Check changes to core sample schema - @daniwelter 
- [x] Check changes to other schema files - @malloryfreeberg 
I think all schema files have been reviewed by me and @daniwelter now. Comments made where questions still exist.


json_schema/sample.json
```diff
@@ -1,70 +1,92 @@
...
+		      "sample_from_organism",
+		      "cell_suspension"
             ]
         },
-        "protocols": {
-            "items": {
-                "$ref": "protocol.json",
-                "description": "An array of protocol modules.  Assay and project protocols reference this."
-            },
-            "type": "array"
```
daniwelter: Why have we removed the protocols link from sample?

json_schema/sample.json
```diff
@@ -1,70 +1,92 @@
...
-                { "$ref": "cell_suspension.json" },
-                { "$ref": "organoid.json" } 
+        "type": {
+            "description": "The name of the core metadata entity type.",
+            "enum": [
+		      "donor",
+		      "sample_from_organism",
+		      "cell_suspension"
             ]
         },
```
daniwelter: Why is there an enum back in here, and where have cell_line and organoid gone?

json_schema/sample_from_organism.json
```diff
@@ -20,10 +16,13 @@
         "state_of_specimen": {
             "$ref": "state_of_specimen.json",
```
daniwelter: Reviewed and agreed.

json_schema/core.json
```diff
@@ -23,7 +22,7 @@
 		      "analysis"
             ]
         },
-        "user_supplied_id":{
+        "id":{
```
daniwelter: Why have we renamed this from user_supplied_id?

json_schema/donor.json
```diff
@@ -128,8 +121,9 @@
         }
     }, 
     "required": [
-        "genus_species", "is_living", "sample_type"
+        "species", "is_living"
```
daniwelter: This should be genus_species @simonjupp

json_schema/ingest.json
```diff
@@ -3,7 +3,7 @@
     "additionalProperties": false,
     "description": "Information added or generated at ingest time." ,
     "properties": {
-        "submission_date": {
+        "submissionDate": {
```
daniwelter: why are we camel-casing dates but underscoring other fields? @simonjupp

json_schema/ingest.json
```diff
@@ -39,6 +34,6 @@
 
     },
     "required": [
-        "submitter_id", "submission_date"
+        "document_id", "submissionDate"
```
daniwelter: Shouldn't the submitter_id be required here as well, as it should be pulled directly from the user login? @simonjupp

src/sample_test.py
```diff
@@ -0,0 +1,29 @@
+from schema_test_suite import get_validator, get_json_from_file, validate
```
dosumis: Happy to leave this for now, but I want to move to a system where each test doesn't need to be added by hand.  e.g. all sample test json files live here:

test_schemas/
.      samples/
.       .        fu.json

Scripts are in place for this in v4 branch.

json_schema/core.json
malloryfreeberg: I thought we agreed to "user_supplied_id".

json_schema/cell_line.json
```diff
@@ -23,23 +23,15 @@
             "description": "when was the cell line established", 
             "format": "date-time",
             "type": "string"
-        }, 
-        "derived_from": {
```
dosumis: Removal of this field makes sense for immortalized cell line (not derived from any sample we care about), but likely to still be required for primary_cell_line.  This was actually one of the sample subschemas we said we'd create.  Ticket: #45

json_schema/cell_line.json
```diff
@@ -23,23 +23,15 @@
...
         }, 
         "karyotype": {
             "description": "the karyotype of the cell line", 
             "type": "string"
-        }, 
-        "name": {
-	        "description": "what is the official name for the cell line?",
-            "$ref": "ontology.json"
-        }, 
+        },
```
malloryfreeberg: `name` should stay (maybe re-named) if we are to link to an ontology for cell lines (which we should).

json_schema/sample_from_organism.json
```diff
@@ -20,10 +16,13 @@
...
             "description": "State of body_part at collection and how it was preserved after removal."
+        },
+        "donor_id" : {
+            "type" : "string",
+            "description" : "the local identifier for the donor of this sample"
```
malloryfreeberg: 👍

json_schema/donor.json
malloryfreeberg: +1 as users are more likely to give us "homo sapiens" if we ask genus_species

json_schema/donor.json
```diff
@@ -63,11 +60,7 @@
             "maximum": 10, 
             "minimum": 0, 
             "type": "number"
-        }, 
-        "id": {
-            "description": "A unique ID for this donor", 
-            "type": "string"
-        },
```
malloryfreeberg: We need a `user_supplied_id` field for donors

json_schema/donor.json
malloryfreeberg: Also, an ID should be required as it will be the way that submitters can link a donor (1) to all it's derived samples (many).

json_schema/sample.json
simonjupp: this can go and needs to be pushed down into the subschemas as type with 1 allowable value for each

json_schema/sample.json
simonjupp: protocols are related through ingest API now so don't think we need this

json_schema/core.json
simonjupp: given the user has to give us this it is implied that it is user supplied. id seems simpler and lets make it clear in the description

json_schema/donor.json
simonjupp: sample already has id and ncbi_taxon_id as required, so donor will inherit these properties

json_schema/ingest.json
simonjupp: this is how it currently is in ingest, I agree and ingest should change, but for now let's keep it like this and create a ticket for ingest dev team to change it

json_schema/ingest.json
simonjupp: agree, but ingest api doesn't provide this yet, so let's leave it off until we can test it against ingest.

src/sample_test.py
simonjupp: yes, leave for now until we have a better replacement

json_schema/cell_line.json
simonjupp: derived_from is reserved for ingest so we need something else for cell lines

PR #63, opened 2017-11-06
Adding back simple schema checks

dosumis:
Adding back test of all schema files for validity as JSON schema.  This may break travis due to problems with analysis.json that we inherited from version_3


malloryfreeberg:
This PR can be merged with version_4 and travis should still pass.

PR #65, opened 2017-11-21
Updated/added id and schema URLs to sample types.

malloryfreeberg:
- Added "id" field to all sample types
- Added "$schema" property to all sample types and populated with regex for json url
- Checked that "sample_core" was in all sample types
- Updated missing descriptions, additionalProperties, title, required for sample types


malloryfreeberg:
@daniwelter If you can quickly review this PR, and Ok it, I'll merge. I want a second set of eyes in case I missed something obvious.

Issue #85, opened 2017-11-30
Subdivision of organism/donor fields based on species


malloryfreeberg:
In line with #83 some users did not want to see e.g. "smoking_history" if they had mouse data.

Suggest logically grouping human and non-human fields as definitions in the organism/donor.json schema to help with visualising only relevant fields.

Issue #84, opened 2017-11-30
Rename "ancestry" field in donor.json


malloryfreeberg:
Users did not know what to put for ancestry field in donor.json, nor did they know what ancestro was.

Suggest changing this field to something like "ethnicity" which is more reflective of what this field is actually asking for.

Issue #83, opened 2017-11-30
Rename "donor" to more generic (e.g. "organism")


malloryfreeberg:
Based on user feedback, some projects use non-human subjects (e.g. mouse, dog), in which case the concept of a "donor" does not apply. Some users were very hesitant to put non-human subjects in the "donor" tab.

Suggest renaming "donor" to something more generic. For example, "organism". Other suggestions?

lauraclarke:
Is there a way to have differential requirements across different species?

malloryfreeberg:
Re-opening this issue as there is now more discussion going on about the use of donor versus organism. Below are quotes from pilot users taken during user experience evaluation sessions and notes from the Boston metadata meeting specifically centered on metadata terms which led to the decision to replace donor with organism as the entity name.

1. Quote from pilot user who has both human and mouse data: "Donor is for the human samples, right? ... [A mouse is] an organism, but not a donor. At least not a voluntary donor."

2. Scribe notes from Boston metadata meeting: "The HCA considered Donor to be the whole organism. Most data contributors put their "specimen" under "donor" and not "specimen from organism". Why? This is because the term "donor" was confusing."

3. Scribe notes from Boston metadata meeting: Data generators thought "Immune cells, brain tissue" were "donor" type samples. They should be "specimen from organism" types.

Another suggested term instead of donor was "individual", but "organism" was chosen instead because even the term "individual" has some suggestion of person-hood.

malloryfreeberg:
Also copying a relevant, recent Slack conversation here between myself and @pnejad:

"Could you explain the reasoning behind not using the term donor?"

"So, given that we are expecting to receive non-human samples, we got some feedback from users that the term "donor" implies a human giving consent, which might not always be the case. So, we switched to the more neutral and general "organism" which removes a sense of the individual giving consent (which non-humans can not give) and also makes clear that this schema is for the "whole individual"."

"In regards to donor and consent, wouldn't the non-consent issue also apply to samples from embryos, fetuses and children? Other species can't consent to their samples being donated to research, but likewise some human samples also don’t have consent from donor."

"Yup! There are many instances where the idea of "consent" doesn’t apply, which is why we decided to drop donor in favor of organism."

malloryfreeberg:
Another point to consider brought up by @claymfischer:

[...]

PR #86, opened 2017-11-30
Updated donor.json to organism.json

malloryfreeberg:
Updated donor.json, including changing to organism.json
- Renamed "ancestry" field in donor.json #84 
- Renamed "donor" to more generic "organism" #83 
- Moved human-specific and non-human-specific fields up to definitions in organism.json #85
-


JimKent:
Good idea.  Organism is a very neutral term.  If there is any problem, it's that some people will associate it with species.  I can't think of anything better though, and this is an improvement over donor.

malloryfreeberg:
Thanks, @JimKent

Issue #14, opened 2017-10-16
analysis.json is not a valid json schema


dosumis:
Tested using python jsonschema:

```
SchemaError: 'string' is not of type 'object'
Failed validating 'type' in schema['properties']['definitions']['additionalProperties']['properties']['properties']['additionalProperties']
On instance['definitions']['parameter']['properties']['name']:
    'string'
```

Looks like there may be a nesting problem with definitions.

dosumis:
Hi @TimothyTickle  - from what I can gather, you might be the best person to fix this.  So have assigned for now.  Cheers, David

daniwelter:
This passes in v4

daniwelter:
analysis.json has been renamed analysis_process.json in v5 and now fully passes validation.

PR #96, opened 2017-12-13
Revised analysis.json so it becomes valid json

daniwelter:
analysis.json has been failing validation as it isn't valid json in its currently release form. I have made the following changes in order to make it valid json:
1. made all properties in the definitions into objects
2. changed all uri types to string
3. changed int types to integer
4. corrected the way "items" with $ref are formatted in properties

@TimothyTickle  I would be grateful if you could please review and merge into master if you're happy with the changes, or assign to another reviewer.

### DIFF
--- a/src/simple_test.py
+++ b/src/simple_test.py
@@ -8,7 +8,7 @@ Specific schema tests to follow.
 """
 
 # a*.json skipped for now
-schemas = glob.glob('../json_schema/' +"[b-z]*.json")
+schemas = glob.glob('../json_schema/' +"[a-z]*.json")
 
 for s in schemas:
     print("Checking whether %s is a valid json schema" % s) 
--- a/json_schema/analysis.json
+++ b/json_schema/analysis.json
@@ -1,154 +1,184 @@
 {
-    "title": "analysis", 
+    "title": "analysis",
     "required": [
-        "timestamp_start_utc", 
-        "timestamp_stop_utc", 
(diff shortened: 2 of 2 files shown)


dosumis:
Ticket here: #14

malloryfreeberg:
I might suggest making `"additionalProperties": false` rather than `true` to be consistent with the rest of the schemas. This change relates to the discussion of how flexible vs rigid we want the schema to be, but unless there is an explicit reason for expecting other properties, I vote for being consistent with this detail for now.

TimothyTickle:
@samanehsan Will be the contact for this update; she will approve the PR or follow-up with feedback. Thank you for your work on this.

TimothyTickle:
Note, as far as the additionalProperties goes, I personally would like a flexible approach to the *.jsons in general (as per our conversation in the metadata meeting last meeting time). That being said, I respect your choice here and value consistency (given technical approval from @samanehsan ).

src/simple_test.py
```diff
@@ -8,7 +8,7 @@
 """
 
 # a*.json skipped for now
```
samanehsan: It looks like you can remove this comment now

daniwelter: Yes, this can go if we get the go ahead on the corrected analysis.json

daniwelter: Ah, thanks, just seen the approval :)

PR #118, opened 2018-01-30
Added a growth_conditions biomaterial module.

malloryfreeberg:
This PR is to add a new module - growth_conditions - to the set of biomaterial modules. This module contains fields relating to the growth or maintenance of a biomaterial in culture. This module has been added to the schemas for cell line, organoid, and cell suspension. The changes include:
- Addition of the module - `growth_conditions.json` - to `json_schema/module/biomaterials/`
- Addition of `growth_conditions` field to `cell_line.json`, `cell_suspension.json`, and `organoid.json`
- Removal of `passage_number` field from `cell_line.json` because it is now in the module
- Added missing `user_friendly` attribute to `cell_morphology` field in `organoid.json`

### DIFF
--- a/json_schema/module/biomaterial/growth_conditions.json
+++ b/json_schema/module/biomaterial/growth_conditions.json
@@ -59,6 +59,12 @@
             "type": "string",
             "example": "No spots of bright blue stain observed at high magnification",
             "user_friendly": "Mycoplasma testing results"
+        },
+        "drug_treatment": {
+            "description": "Description of drugs added to the growth medium for this biomaterial.",
+            "type": "string",
+            "example": "100 ug/mL ampicillin",
+            "user_friendly": "Drug treatment"
         }
     }
 }
--- a/json_schema/module/biomaterial/growth_conditions.json
+++ b/json_schema/module/biomaterial/growth_conditions.json
@@ -1,0 +1,64 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "id": "http://schema.humancellatlas.org/module/biomaterial/5.0.0/growth_conditions.json",
+    "description": "Information relating to how a biomaterial was grown and/or maintained in a laboratory setting.",
+    "additionalProperties": false,
+    "required": [
+        "$schema"
+    ],
+    "title": "growth_conditions",
+    "type": "object",
+    "properties": {
+        "$schema" : {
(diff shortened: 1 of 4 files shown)


JimKent:
I had a look and generally it works.  Passage number doesn't seem really relevant to cell_suspension or organoids though.  Also is growth condition where you're thinking to put drug treatments?

malloryfreeberg:
Thanks for the feedback, Jim. 

Regarding drug treatments, can you clarify a specific example of a drug treatment that would be appropriate here? I'm trying to think about how to model this metadata. Would drugs sometimes be included in the growth medium. The only other field we have relating to drug treatment is in the medical_history module (which only hangs off the organism type at the moment), where we have a medication field and a treatments field. But, I can imagine knowing whether drugs were used in the growth of cells is important.

Alternatively, feel free to add a drug_treatment field to the schema directly, and I'm happy to review/merge it.

JimKent:
Yes, typically the drugs are added to the medium.  In 90% of cases the cells need to be alive to respond to the drugs.  A common example would be adding interleukins or other cytokines to a culture of immune cells to activate them so we get the genes that are happening in response to infection rather than the minimal transcription of a resting lymphocyte.

JimKent:
So "culture_conditions" might be more accurate than "growth_conditions" as a module name, though growth_conditions is not bad.

malloryfreeberg:
I added a free text field - `drug_treatment` - to capture drugs added during the growth process. 

I definitely am aware that we need to be careful and avoid making the schema too granular; however, I also want to include, up-front, fields that represent metadata that users are actively asking to submit values for. This approach has the benefit of decreasing the curation cost downstream, when our ingest load become unmanageable for one-on-one broker-submitter interactions. So, I am a big fan of the bottom-up approach. 

I regret that there is not a great way (that I've found) to link our schema changes to feedback we get from UX sessions and pilot submitters (most of which is in Google Drive), but I will try to make it more clear that many of the content additions/changes are driven by user feedback. In the case of this module, all the included fields were either present in v4.6.1 or were direct suggestions from the metadata community meeting we had at the Broad in January (and subsequently discussed amongst other DCP members). 

Thanks for all of the feedback. I'm definitely glad to be making this process as transparent as possible.

JimKent:
Thanks all.

Issue #123, opened 2018-01-31
Remove "name" from researcher field in process_core


malloryfreeberg:
Replace with something like "operator identity". We can't ask for names as this is identifying information.

Issue #92, opened 2017-12-07
Why does the library_construction field use enums?


malloryfreeberg:
Why are the library_construction fields in rna.json and seq.json enums? I've had to add to these lists to ingest pilot data because "10x" wasn't an option. I'm also not entirely sure what the difference is between the fields - should they be the same? If so, do we need both? The descriptions are unclear to me:

rna.json: 
```
"description": "The general approach for sequencing library construction. e.g. Smart-seq, Drop-seq, 10x.",
"enum": [
                "unknown", 
                "CEL-seq", 
                "SMARTer Ultra Low RNA Kit", 
                "modified smart-seq2", 
                "smart-seq2", 
                "QUARTZ-Seq", 
                "10x_v2", 
                "drop-seq", 
                "inDrop", 
                "10x", 
                "10X"
], 
```

seq.json:
```
"description": "How the DNA sequencing library was prepared. e.g. Nextera XT, TrueSeq."
"enum": [
                "Nextera XT", 
                "TruSeq", 
                "modified Nextera XT", 
                "10X", 
                "10x"
], 
```

Unless there is a compelling reason to make these fields enums, would it be better to just make them strings to avoid restricting users? It's quite a pain to enforce matching to all these enums in light of things like capitalization, punctuations (drop-seq vs DropSeq), etc.

dosumis:
Technique used should really be recorded using EFO.
On Dec 7, 2017 4:12 PM, "Mallory Freeberg" <notifications@github.com> wrote:

Why are the library_construction fields in rna.json and seq.json enums?
I've had to add to these lists to ingest pilot data because "10x" wasn't an
option. I'm also not entirely sure what the difference is between the
fields - should they be the same? If so, do we need both? The descriptions
are unclear to me:

rna.json:

"description": "The general approach for sequencing library
construction. e.g. Smart-seq, Drop-seq, 10x.",
"enum": [
                "unknown",
                "CEL-seq",
                "SMARTer Ultra Low RNA Kit",
                "modified smart-seq2",
                "smart-seq2",
                "QUARTZ-Seq",
                "10x_v2",
                "drop-seq",
                "inDrop",
                "10x",
                "10X"
],

seq.json:

"description": "How the DNA sequencing library was prepared. e.g.
Nextera XT, TrueSeq."
"enum": [
                "Nextera XT",
                "TruSeq",
                "modified Nextera XT",
                "10X",
                "10x"
],

[...]

PR #130, opened 2018-01-31
V5 prototype schema reviews

### LINKED ISSUE #81: Height in cm instead of m
User suggests recording height in centimeters instead of meters as this is how it typically comes from the clinic.

### LINKED ISSUE #125: Change sex to biological_sex
In organism.json schema

### LINKED ISSUE #126: Add height and weight unit fields
To organism.json

### LINKED ISSUE #127: Make total_estimated_cells an integer
Total_estimated_cells field in cell_suspension type is a number ranging from 0.1 to 1000000000.0. This should be an integer.

### LINKED ISSUE #128: Comment fields
- library_construction: will become an ontology

### LINKED ISSUE #129: Missing medical_history field in organism.json
Organism.json needs to reference the medical_history module.

malloryfreeberg:
Minor revisions made based on a global review of all core and type fields in all schemata. Revisions include:
- Changed `name` to `operator_identity` to describe person who did a process. Fixes #123. 
- Changed `sex` to `biological_sex`. Fixes #125.
- Added unit fields for height and weight in `organism.json`. Fixes #126. Fixes #81.
- Changed `total_estimated_cells` from number to integer. Fixes #127.
- Added comment that `library_construction` will become an ontology. Fixes #128. Also referenced in #92.
- Added `medical_history` field to reference module in `organism.json`. Fixes #129.

### DIFF
--- a/json_schema/core/process/process_core.json
+++ b/json_schema/core/process/process_core.json
@@ -65,7 +65,7 @@
             "user_friendly": "Location"
          },
          "operator_identity": {
-           "description": "Identity of individual(s) who executed this process.",
+           "description": "Identifier for individual(s) who executed this process.",
            "type": "array",
            "example": "Technician 1",
            "items":{
--- a/json_schema/core/process/process_core.json
+++ b/json_schema/core/process/process_core.json
@@ -64,14 +64,14 @@
             "example": "Wellcome Sanger Institute, Teichman Lab.",
             "user_friendly": "Location"
          },
-         "researcher": {
-           "description": "Name of the individual(s) who executed this process.",
+         "operator_identity": {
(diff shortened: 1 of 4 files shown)


json_schema/core/process/process_core.json
```diff
@@ -64,14 +64,14 @@
             "example": "Wellcome Sanger Institute, Teichman Lab.",
             "user_friendly": "Location"
          },
-         "researcher": {
-           "description": "Name of the individual(s) who executed this process.",
+         "operator_identity": {
+           "description": "Identity of individual(s) who executed this process.",
```
daniwelter: I would phrase this "identifier for.." rather than identity, specifically to make it clear that we don't want their identity

PR #141, opened 2018-02-01
V5 prototype new preservation/storage module

malloryfreeberg:
This new module contains information that was actually lost from v3, information about how a specimen was preserved and/or stored. Preservation specifically refers to the use of chemicals, cold, or other means to prevent or retard biological or physical deterioration of a biomaterial. Storage specifically refers to how a biomaterial was stored after preservation or before another protocol was used. This module currently is associated with the specimen_from_organism entity, but it is generic enough that we might consider allowing it to be submitted with cell_suspension, cell_line, and organoid. It is probably not relevant for organism.


JimKent:
Looks good.  I'm going ahead and merging.

PR #147, opened 2018-02-06
V5 prototype new process bundle

malloryfreeberg:
This PR includes changes to update to a new process_bundle schema from the separate assay and analysis bundle schemas. Specifically, changes include: 
- Deleted analysis_bundle.json
- Made process_bundle.json and protocol_bundle.json
- Added `oneOf` for content $refs for process_bundle.json
- Added valid example process_bundle JSON to validate against its schema

Changes to biomaterial_bundle:
- Changed derived_from to array because a biomaterial may be pooled
- Changed derivation_protocols to derivation_processes because biomaterials will be linked by processes not protocols

General:
- Started including only the appropriate enum value in the `schema_type` field


malloryfreeberg:
@daniwelter can you review and merge this? I know it's failing Travis but this is due to the refs resolving against v5_prototype instead of this branch (where I've made the changes). After the merge I'll clean up any remaining unresolved things. I know this isn't the best way to fix this, but until we can implement a better way, it's easier for me to see where all the linkings are failing post-merge. Thanks!

PR #151, opened 2018-02-07
Ontology module updates

daniwelter:
Added normal as property option to disease_ontology module, added mass_unit_ontology module and updated height/weight unit in organism to point to relevant ontology modules


malloryfreeberg:
Keeping this branch open for future ontology module updates.

daniwelter:
I just deleted the branch - its naming was specific to the disease ontology module, so other modules should be covered by a more generic branch

PR #153, opened 2018-02-07
Schema type updates

daniwelter:
Restricted schema_type enum to a single value for all core and type entities, and to one or more for module entities as required. Removed schema_type from ontology module entities as not appropriate here


malloryfreeberg:
Double check that Travis passes after merge.

PR #158, opened 2018-02-07
Added user friendly term to familial_relationship in organism.json.

pnejad:
Added user friendly term to familial_relationship in organism.json.


malloryfreeberg:
Once @cjvillar has approved this change, @pnejad can merge!

PR #159, opened 2018-02-07
added example for organism_age_unit in type/biomaterial/organism.json

willrockout:


willrockout:
@malloryfreeberg I guess i'm a little confused on what you mean. Can you explain a bit I get that organism_age_unit is references the time_unit_ontology.json but i'm confused on the mapped ontology part. I see in the file under the ontology term you have ontologies with prefix and identifier (obo:efo and obo:uo) are you asking one of these but in reference to year?

willrockout:
if thats the case then would it be UO:0000036?

malloryfreeberg:
Yes! That's exactly what I meant. I wanted to ensure there is awareness that our examples should definitely be things that would be valid in the schema. So it's good practice to be sure the example ontology term we provide is indeed in the ontology.

malloryfreeberg:
@willrockout go ahead and merge/delete when ready

willrockout:
@malloryfreeberg Quick question should i be adding that to the time_unit_ontology.json also?

malloryfreeberg:
It would be great if you could! Adding examples is fairly low priority in terms of getting a complete v5 out, but it's good experience so I'm not going to stop you if you want to :) You can commit/push to this same branch (dont merge yet) and then ask for another review when you have made all the changes.

willrockout:
@malloryfreeberg Hey can you check and see what the problem is not sure if i have to add to ontologies in that ontology or can you just add classes? Also in the code not sure whats with the added newline i've tried removing it in code and on github and it won't go away

malloryfreeberg:
Two things.
1. I misunderstood your previous question. You do not need to actually add the ontology term UO:0000036 to `time_unit_ontology.json` if the term lives under one of the terms already listed (which I think it does). So, there shouldn't be a need to change this particular ontology module.
2. The problem with the automated testing is that now the v5_prototype branch - which is the branch you originally branched off of for your working v5_prototype_hw branch - is a few commits ahead of when you branched off of it and has important changes that you need to bring into your branch. To bring your v5_prototype_hw up to the current v5_prototype, you just need to do `git rebase v5_prototype` while in your v5_prototype_hw branch, and then the tests should run. 

So, I would remove the "UO:0000036" term from the `time_unit_ontology.json` file, rebase to the current v5_prototype branch, re-run the test, and everything should be good to go :) I can review again after you make these changes.

willrockout:
@malloryfreeberg should be good now I removed the ontology term so its just my example

PR #167, opened 2018-02-08
Add disease field for specimen. Fixes #11.

### LINKED ISSUE #11: Need a field that indicates tissue, as opposed to patient, is diseased.
This example shows why:

```json
{ "short_label" : "U2OS Cells", 
"long_label" : "Osteosarcoma from the tibia of a 15 year old", 
"uuid" : "5f16b743defc4b5c8717f023f9921069", 
"body_part" : 
	{ "name" : "tibia" }, 
"donor" : 
	{ "id": "U2", 
	"uuid": "a1ef5009045f42e88a0e3db1d1f2b050", 
	"life_stage" : "child", "age": 15, "species": "Homo sapiens" } }
```

Formal metadata makes this sample look like normal body part, rather than a tumor sample!
Asserting that a sample exhibits a disease is distinct from asserting that a patient does.  A sample may come from a normal bone from a patient with osteosarcoma.

This is an action item from the DCP meeting at CZI:
https://docs.google.com/document/d/1w-OKngEuIC0oYEdg_QtaIZFQ8BSAZQIJR3qumSBL_ho/edit#

malloryfreeberg:
We need to allow users to declare that their sample represents a certain disease separate from a disease the organism might have that may or may not be related to the disease of the specimen.

### DIFF
--- a/json_schema/type/biomaterial/specimen_from_organism.json
+++ b/json_schema/type/biomaterial/specimen_from_organism.json
@@ -58,6 +58,14 @@
             "comment": "The term organ part is very broadly defined here.",
             "user_friendly": "Organ part"
         },
+        "disease": {
+            "description": "Short description of disease status of the specimen.",
+            "type": "array",
+            "items": {
+                "$ref": "https://raw.githubusercontent.com/HumanCellAtlas/metadata-schema/v5_prototype/json_schema/module/ontology/disease_ontology.json"
+	        },
+            "user_friendly": "Disease"
+        },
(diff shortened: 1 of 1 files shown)


lauraclarke:
surely it can't have must be. If someone doesn't know if the person is "normal" or diseased they would want to leave it blank to indicate lack of knowledge

malloryfreeberg:
Well it's not required. I guess "must be" is strong language.

daniwelter:
What about "Should be"?

PR #175, opened 2018-02-09
Removed .json extension from all ID and  fields

daniwelter:
I know tests are currently failing but that's because it's looking at the online v5_prototype json files, which still have .json in the regex pattern. Tests should pass straight after the merge.


malloryfreeberg:
Merging with understanding that @daniwelter will check whether Travis CI tests pass once changes are a part of v5_prototype branch.

PR #205, opened 2018-02-20
Add new amplification_method field to lib prep. Fixes #197.

### LINKED ISSUE #197: Capturing metadata related to amplification methods used
> Q: Is it possible to add amplification method into metadata data? The option can be PCR, IVT, or none. The 10X and SS2 are both PCR amplification but there are also some scRNA-Seq use IVT as amplification method. (@jishuxu)

Suggestion to add an `amplification_method` field as an enum. This field should go in the `library_preparation_process.json` schema.

schema to change: library_preparation_process.json
field name: amplification_method
description: The method by which oligonucleotides are amplified prior to sequencing. Must be one of PCR, IVT, or none.
enum values: PCR, IVT, none
user-friendly name: Amplification method
required: no

malloryfreeberg:
This PR is to add a new optional field - amplification method - to the library preparation schema. It was suggested by a member of the secondary analysis team @jishuxu. It addresses issue #197 .

This change is a minor version update to library_preparation_process.json from 5.0.0 to 5.1.0.

This change is medium acceptance, requires 2 reviewers, and will be open to feedback/comments for 5 working days.


lauraclarke:
@daniwelter This wasn't meant to be immeditately approved.

The lifecycle process is 

This change is a minor version update to library_preparation_process.json from 5.0.0 to 5.1.0.
This change is medium acceptance, requires 2 reviewers, and will take no more than 5 working days.

We can change the lifecycle process if it doesn't work but we should try and operate it first.

daniwelter:
@lauraclarke So what's the actual physical process in Github? I assumed it was reviewers separately giving approval, and if there are two lots of approval, the change is approved formally by merging the PR. Apologies if I misunderstood that.

lauraclarke:
I don't think a PR should be accepted before it is accepted by the reviewers surely? as the PR acceptance should trigger the new release of the schema 

I am not sure what the best git process is, maybe an issue which gets +1s from all assigned reviewers and then the PR gets accepted on that basis.

daniwelter:
In Github terms, PR acceptance doesn't trigger anything as far as I know. The merging of the changes into master is what triggers the release and this PR is only for merging into develop. Once that's done, we can update the relevant version numbers, then do the merge to master to release. In theory, we could do version number upgrades as part of this PR but that's not helpful if we want to batch multiple changes into one release.

lauraclarke:
Ah, okay

Sounds like we need to discuss my vision for the process from a social sense and line it up with what is possible in git from a technical point of view

lauraclarke:
The other thing to consider if the social view of someone who isn't familiar with git, if they make a PR (or a wranger makes a PR for them) and they see it is approved, assumptions might be made about what it means

json_schema/type/process/sequencing/library_preparation_process.json
```diff
@@ -133,6 +133,16 @@
             "type": "object",
             "$ref": "https://schema.humancellatlas.org/module/process/sequencing/5.0.0/barcode"
         },
+        "amplification_method": {
+            "description": "The method by which oligonucleotides are amplified prior to sequencing. Must be one of PCR, IVT, or none.",
```
jishuxu: More precisely:"The method to amplify cDNA library prior to sequencing, Must be PCR, IVT or none."

malloryfreeberg:
Thanks @jishuxu ! Made the change.

json_schema/type/process/sequencing/library_preparation_process.json
```diff
@@ -133,6 +133,16 @@
...
         },
+        "amplification_method": {
+            "description": "The method to amplify cDNA library prior to sequencing. Must be one of PCR, IVT, or none.",
+            "type": "string",
+            "enum": [
+                "PCR",
+                "IVT",
+                "none"
+            ],
+            "user_friendly": "Amplification method"
```
ambrosejcarr: This could be `[cDNA | Library] amplification method`. Either word would work, both would provide a bit more detail for the user friendly description. Depending on how you guys feel about the below question, perhaps it would make sense to make this `Library pre-amplification method`. 

Follow-up question: 
Most sequencing approaches have multiple rounds of amplification, and those are not always the same. I checked with Vaidas Kiseliovas in Dana Pe'er's lab and he confirmed that almost all protocols follow a 2 step amplification scheme. 

Paraphrasing Vaidas, the reason is following: in the first step you are amplifying full length transcripts, the you have to do shearing, adapter ligation into smaller fragments and adapter ligation, then you do additional PCR to enrich for fragments containing both adapters for sequencing. The only exemption we are aware of is ddSEQ (Drop-Seq by Illumina and Biorad), they employ one step amplification as their transposase is doing both steps at the same time, fragmentation and ligation followed by single PCR step to enrich for fragments with both adapters.

Some InDrops approaches use IVT for the first round and PCR for the second round. Additionally, I believe *most* IVT protocols would probably use PCR for the second cycle, as using IVT would force several follow-up enzymatic reactions to get cDNA back. 

Based on this information, should we support two different amplification approaches? Should we expect `IVT` to mean that there is always a second PCR amplification round?

jishuxu: Yes, I think this schema is for library amplification, not pre-amp. it should be applied to cDNA library. The pre-amplification is kind of complicated for full length library, different protocols use different pre-amp method.

malloryfreeberg: This is a good distinction to know, thanks for the feedback! It sounds like it will be important to capture both amplification methods as both steps have the potential to introduce bias. If that's the case, what should the accepted values and description be for each? I've taken a stab at it:

Field 1: Library pre-amplification method
Description: The method to amplify RNA prior to adaptor ligation. Must be one of: <values>. 
Values: PCR, IVT, none

Field 2: cDNA library amplification method
Description: The method to amplify cDNA library prior to sequencing. Must be one of: <values>.
Values: PCR, none

malloryfreeberg:
@ambrosejcarr @jishuxu How do the following two new fields look?

Field 1: Library pre-amplification method (`library_preamplification_method`)
Description: The method used to amplify RNA prior to adaptor ligation. Must be one of: (values).
Values: PCR, IVT, none

Field 2: cDNA library amplification method (`cdna_library_amplification_method`)
Description: The method used to amplify a cDNA library prior to sequencing. Must be one of: (values) .
Values: PCR, none

malloryfreeberg:
Update: We would like to make these two fields ontologies so that the methods are standardized. We want to avoid simply making this an enum (because that is harder to maintain and will require lots of versioning). The EFO ontology is being upgraded after Easter, and we can link to the appropriate terms at that time.

malloryfreeberg:
@daniwelter Can you determine whether these fields can be ontologized with at least PCR and IVT (in vitro transcription)?

malloryfreeberg:
@daniwelter can you merge pwease?

PR #214, opened 2018-02-22
Rename schema to describedBy

daniwelter:
The decision was made to rename schema to describedBy to conform with best practice in json schema


daniwelter:
This PR also includes changes to the bundle schema structure.

malloryfreeberg:
Here is a [reference](https://tools.ietf.org/id/draft-handrews-json-schema-00.html#rfc.section.11.1) to the correct usage of `describedBy`. From that link: "It is RECOMMENDED that instances described by a schema provide a link to a downloadable JSON Schema using the link relation "describedby", as defined by Linked Data Protocol 1.0, section 8.1."

Issue #217, opened 2018-02-27
"title" should be "publication_title"


Bento007:
It looks like a bug, please correct me if i'm wrong. This https://github.com/HumanCellAtlas/metadata-schema/blob/290696a0fa77a777b085f94d534f19c79de4d628/json_schema/module/project/publication.json#L9 should be https://github.com/HumanCellAtlas/metadata-schema/blob/290696a0fa77a777b085f94d534f19c79de4d628/json_schema/module/project/publication.json#L34

malloryfreeberg:
Thanks for the catch, @Bento007 ! A bug fix patch has been released.

PR #218, opened 2018-02-28
Malloryfreeberg bug fix issue217

malloryfreeberg:
As referenced in Issue #217, `title` should be `publication_title` in the list of required fields in the `publication` module. This change is a patch because it is a bug fix. References to this module in `cell_line.json` and `project.json` were also update to reflect the version change.


daniwelter:
Version number updates reviewed in all files and they look correct.

PR #221, opened 2018-03-01
Added glio and pbmc filled spreadsheet examples

willrockout:
Hey i added the two example spreadsheets


malloryfreeberg:
@willrockout I changed the base branch to develop. All changes (even ones not to metadata schema) should go through the develop branch. The ensures that we have a second layer of security in case something weird/bad gets merged into master and also allows the develop branch to always be up-to-date with master, which is important because we want all PRs and suggested changes to be based on develop. I'll approve and @daniwelter can merge tomorrow morning.

Issue #208, opened 2018-02-21
F2F Meeting: Resource Bundle Metadata Schema


kbergin:
Aims: Working session to choose metadata for a reference bundle and check any needs by the blue box for indexing.

Outcomes: a PR for resource metadata schema

We will need representation from green on engineering and pipeline side, and from blue box, someone familiar with indexing and the communication between green and blue.

malloryfreeberg:
Since this will be a working session to produce a PR for adding a new schema, I suggest the following roles for this session:

**Co-leads - lead discussion, present requirements for reference bundle, :**
Dani @daniwelter (ingest)
Jishu @jishuxu (pipelines)
Trent @Bento007 (indexing)

**Hacking - make reference bundle JSON schema that meets the needs of all teams, open a PR:**
Mallory (ingest)

At the end of the session, I will make a PR and the metadata team will follow the update process for incorporating new metadata schema.

malloryfreeberg:
Closing this issue as it was addressed by commit 95fea68.

PR #227, opened 2018-03-06
AI from reference bundle breakout 

daniwelter:
We created reference_file.json and reference.json bundle plus bundle example. Addressing github issue #208. Please review and comment by 20 March 2018. This PR requires approval by at least two independent experts.


json_schema/type/file/reference_file.json
```diff
@@ -0,0 +1,75 @@
...
+            "description": "Core file-level information.",
+            "type": "object",
+            "$ref": "https://schema.humancellatlas.org/core/file/5.0.0/file_core"
+        },
+        "ncbi_taxon_id" : {
+            "description": "A taxonomy ID (taxonID) from NCBI.",
+            "type": "integer",
+            "user_friendly": "NCBI taxon ID"
+        },
+        "genus_species": {
```
dshiga: This seems redundant with the taxon id above, but I'm guessing there's a reason why you've included it here -- can you explain more?

schema_test_files/file/test_pass_reference_bundle_0.json
```diff
@@ -0,0 +1,121 @@
+{
```
dshiga: Is this a copy of examples/bundles/v5/reference.json? Is there a reason to keep both or should this be consolidated?

json_schema/type/file/reference_file.json
```diff
@@ -0,0 +1,75 @@
...
+            "type": "string",
+            "enum": [
+                "genome sequence",
+                "transcriptome sequence",
+                "annotation reference",
+                "transcriptome index",
+                "genome sequence index"
+            ]
+        },
+        "reference_version": {
```
dshiga: Are there any constraints here? For example people could create inconsistent bundles where one says "GencodeV27" and another says "GENCODE v27"? Do we want constraints or are we okay with letting reference bundle submitters be responsible for choosing these strings in a consistent way?

json_schema/type/file/reference_file.json
malloryfreeberg: The reason for having both `ncbi_taxon_id` and `genus_species` fields among all the biomaterial entities is to have an option for users to give us either the taxon ID or the latin species name, depending on what they are comfortable with. We therefore included both fields in the `reference_file.json` entity to be consistent with the biomaterial types. Neither fields are required, and removing schema fields is a breaking change, so I'd rather err on the side of anticipating we might need both in the future and therefore keep them in this schema to maintain consistency.

schema_test_files/file/test_pass_reference_bundle_0.json
malloryfreeberg: We have been reserving `schema_test_files` for containing JSON files that are used during our Travis CI tests, and `examples` for example JSON, spreadsheets, etc. that we get requests for from various groups/boxes. Maybe @daniwelter can confirm why there is a copy in both places, but my guess is to keep separate examples for other users and the JSON files we use for testing.

json_schema/type/file/reference_file.json
malloryfreeberg: Happy to discuss whether we want restrictions -- either via a regex or an enum. My understanding was that the reference bundles would always be coming from green box - I think @jishuxu mentioned this was because green box needs to run programs to generate the indexes, and so then those results + reference files would be pushed from green to purple to blue. So, we can constrain the reference schemas anyway you want :)

json_schema/bundle/reference.json
```diff
@@ -0,0 +1,59 @@
...
+        },
+        "references": {
+            "description": "An array of reference files.",
+            "type": "array",
+            "items": {
+                "$ref": "https://schema.humancellatlas.org/bundle/1.0.0/reference#/definitions/reference_ingest"
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
malloryfreeberg: Fixed!

json_schema/bundle/reference.json
```diff
@@ -0,0 +1,59 @@
...
+                "content": {
+                    "description": "Content for any reference type entity.",
+                    "type": "object",
+                    "$ref": "https://schema.humancellatlas.org/type/file/5.0.0/reference_file"
+                }
+            }
+        }
+    },
+    "properties": {
+        "describedBy" : {
```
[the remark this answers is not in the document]
malloryfreeberg: Noted. We had this in v4 (called the `core.json` schema) but after moving to v5, we now don't have `describedBy` in core and modules entities. So I'll make a note and then we can reassess (probably after the end of the quarter). Thanks for the suggestion!

json_schema/type/file/reference_file.json
jishuxu: the references+reference indexing files will be generated from greenbox for sure. There are at least two usages for this reference bundle json. 1) we will have another pipeline to generate reference bundles and the reference.json or something like it would be the inputs for this pipeline. 2) we need a metadata to keep tracking the reference we are using in pipeline.

PR #230, opened 2018-03-07
Created a new post-analysis bundle example with v5 metadata.

malloryfreeberg:
In order to help green box migrate their subscription queries over to v5 of the metadata (in which they need to query on presence/absence of an analysis process in the bundle), this PR adds a post-analysis bundle example based on the same SS2 example bundle already in the repo. All of the files in this PR are the same as in `examples/bundles/v5/Q4DemoSS2/` except `process.json` also now contains a block of `analysis_process.json` content.

### DIFF
--- a/examples/bundles/v5/Q4DemoSS2PostAnalysis/biomaterial.json
+++ b/examples/bundles/v5/Q4DemoSS2PostAnalysis/biomaterial.json
@@ -1,0 +1,85 @@
+{
+  "describedBy": "https://schema.humancellatlas.org/bundle/5.0.0/biomaterial",
+  "schema_version": "5.0.0",
+  "schema_type": "biomaterial_bundle",
+  "biomaterials": [
+    {
+      "content": {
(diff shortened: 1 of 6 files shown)


daniwelter:
For the sake of consistency, shouldn't analysis files be added to files.json?

malloryfreeberg:
I wondered about that, but don't know if Green Box is supposed to update/change the pre-analysis bundle. I could be wrong. This inconsistency is a result of changing files produced by secondary analysis to point to a file entity instead of using a definition. If secondary analysis is allowed to change other entities, I'm ok with moving the files they produce to file.json. But, this would be a major version change to analysis_process.json, and green box would have to be aware that they are writing to two entities now (process.json and file.json).

Thoughts? @rexwangcc @samanehsan

daniwelter:
There is a ticket to review how files are handled in analysis.json anyway, which came out of the analysis breakout yesterday. Why don't we accept the current PR for now and then do a wider review of file handling in analysis metadata?

PR #247, opened 2018-03-14
biological_macromolecule_ontology.json has badly formatted array for ontologies

simonjupp:
Ontologies should be a list, but it was a list inside a single string. 

Detected when programming against the schema 

Fixed the formatting of the allowable ontologies for the biological_macromolecule_ontology.json schema


malloryfreeberg:
@simonjupp This would be a bug fix, right?

PR #263, opened 2018-04-19
Simplification suggestions for external facing doc

hewgreen:

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -8,7 +8,7 @@
 
 ## Introduction
 
-This **contributing** document provides directions for suggesting or requesting changes, making pull requests, and reporting bugs in the metadata schema via GitHub. This document also outlines what contributors can expect in terms of acceptance or rejection of contributions. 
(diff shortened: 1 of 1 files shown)


lauraclarke:
Just made a quick scan. We need to be careful with the use of the word broker. This has meaning in software terms and person terms so it would be good to be more specific

Is there are the reason it can't be left as the committer and the valid committers be listed as the wrangling team, then if that gets extended over time we don't need a new term?

hewgreen:
My only point was that committer vs contributor is quite confusing from the outside. Especially in the context of us gradually allowing contributors to commit in a self service system. I just wanted to be clear what steps they can do themselves and what they need to contact us for. Would replacing 'broker' with 'wrangler', 'wrangling team' or 'human broker' solve the problem you highlight? I think Gabby uses the term wrangler but Mallory doesn't think it reflects the task we are performing.

malloryfreeberg:
Closing the PR as it came from Matt's fork, not from a branch in the metadata-schema repo. Matt will re-open a new PR from a metadata-schema repo branch.

PR #268, opened 2018-05-01
Removal of ID field and absolute $ref URLs, addition of version tracking file

daniwelter:
The ID field was removed from all json metadata schemas, and all $ref URLs were changed from absolute URLs for schema.humancellatlas to paths relative to the json_schema directory structure. 
I added a versions.json file which will be used to track version number updates before a release, then used in the release process to insert the correct version number into each ID field.
Some of the test suite had to be disabled (bundles + analysis_process) as the json schema validator doesn't work with multi-level references (eg bundle > type > core/module) - during the second "hop" the base path gets replaced with the path to the 2nd entity, making the next reference resolve incorrectly and causing a test failure due to a FileNotFound exception.


lauraclarke:
I am very concerned if we removing a field which is expected as part of standard JSON schema.

malloryfreeberg:
> in the release process to insert the correct version number into each ID field.

I think @daniwelter can clarify, but the '$id' field is added back in as part of the official release process. Removing '$id' from the GitHub schemas is a workaround for failing validation tests.

lauraclarke:
It would be good to discuss this with people at CZI as the github not having valid schemas feels wrong but I appreciate there may not be another solution

dosumis:
My understanding: 

$id is overloaded in json_schema (see https://github.com/json-schema/json-schema/wiki/The-%22id%22-conundrum).  Given the JSON schema spec, you can either have dynamic (unit) testing of the HCA metadata schema as it evolves (travis build fails whenever a commit breaks the schema) OR you can have $id in the working version.  You can't have both.

lauraclarke:
Okay, so long as it is a known issue with no workaround in the community that is fine. I just wanted to make sure we weren't going to far off track from following standards perspective

malloryfreeberg:
Happy to merge unless we want additional feedback from any third parties.

PR #269, opened 2018-05-01
Adding differentation and ipsc induction processes to type/process/

willrockout:
This update adds process information in type/process/biomaterial_collection/ for iPSC induction and stem cell differentiation. It also adds a module of timecouse information.

### DIFF
--- a/json_schema/type/process/biomaterial_collection/differentiation_process.json
+++ b/json_schema/type/process/biomaterial_collection/differentiation_process.json
@@ -36,12 +36,6 @@
             "type": "object",
             "$ref": "https://schema.humancellatlas.org/core/process/5.1.0/process_core"
         },
-	"differentiation_protocol":{
-	    "description": "Unique identifier of the protocol file which captures the details of how differentiation of cells were performed.",
-	    "type": "string",
-	    "user_friendly": "Differentiation Protocol",
-	    "example": "Abcam- Human embryonic stem cells to hepatic endoderm protocol"
-	},
 	"differentiation_method": {
 	    "decription": "Differentiation method applied to stem cell culture to induce a specific differentiation response.",
 	    "type": "string",
--- a/json_schema/type/process/biomaterial_collection/differentiation_process.json
+++ b/json_schema/type/process/biomaterial_collection/differentiation_process.json
@@ -1,0 +1,134 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "id": "https://schema.humancellatlas.org/type/process/biomaterial_collection/1.0.0/differentiation_process",
+    "description": "Contains information on how a pluripotent cell is differentiated to a desired cell type or organoid.",
+    "additionalProperties": false,
+    "required": [
+        "describedBy",
+        "schema_type",
+        "process_core",
+        "differentiation_method",
+	"differentiation_cell_source_type",
+	"differentiation_target_cell_type"
+    ],
+    "title": "differentiation_process",
+    "type": "object",
+    "properties": {
+        "describedBy":  {
+            "description": "The URL reference to the schema.",
+            "type": "string",
+            "pattern" : "https://schema.humancellatlas.org/type/process/biomaterial_collection/[0-9]{1,}.[0-9]{1,}.[0-9]{1,}/differentiation_process"
+        },
+        "schema_version": {
+            "description": "The version number of the schema in major.minor.patch format.",
+            "type": "string",
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+        "schema_type": {
+            "description": "The type of the metadata schema entity.",
+            "type": "string",
+            "enum": [
+                "process"
+            ]
+        },
+        "process_core" : {
+            "description": "Core process-level information.",
+            "type": "object",
+            "$ref": "https://schema.humancellatlas.org/core/process/5.1.0/process_core"
+        },
+	"differentiation_protocol":{
+	    "description": "Unique identifier of the protocol file which captures the details of how differentiation of cells were performed.",
+	    "type": "string",
+	    "user_friendly": "Differentiation Protocol",
+	    "example": "Abcam- Human embryonic stem cells to hepatic endoderm protocol"
+	},
+	"differentiation_method": {
+	    "decription": "Differentiation method applied to stem cell culture to induce a specific differentiation response.",
+	    "type": "string",
+	    "usert_friendly": "Differentiation Method",
+	    "example": "Embryoid Body, Monolayer, Inductive Co-Culture"
+	},
+	"differentiation_cell_source_id":{
+	    "description": "Unique identifier of the cell source prior to differentiation to target cell.",
+	    "type": "string",
+	    "user_friendly": "Cell Source ID"
+	},
+	"differentiation_cell_source_type":{
+            "description": "Cell type of sample prior to differentiation to target cell.",
+            "type": "string",
+            "user_friendly": "Cell Source Type"
+	    "example": "hESC, iPSC, Mesoderm Progenitor, Neural Progenitor"
+        },
+	"differentiation_target_cell_type":{
+	    "description": "Target cell type that cell of origin is differentiated to.",
+	    "type": "string",
+	    "user_friendly": "Target Cell Type",
+	    "example": "Cardiomyocyte, Cortical Neuron, Macrophage"
+	},
+	"differentiation_target_cell_yield":{
+            "description": "Percent of target cells obtained after directed differentiation of origin cell.",
+            "type": "string",
+            "user_friendly": "Target Cell Yield"
+        },
+	"differentiation_growth_environment":{
+	    "description": "Growth environment in which stem cells are differentiated.",
+	    "type": "string",
+	    "user_friendly": "Growth Environment",
+	    "example": "scaffold, monolayer, suspension, hanging drops, collagen, engraftment"
+	},
+	"differentiation_growth_factors":{
+	    "description": "Growth factor added to stem cell culture to induce a specific differentiation response.",
+	    "type": "string",
+	    "user_friendly": "Differentiation Growth Factors",
+	    "example": "FLT3 Ligand CTS Recombinant Human Protein"
+	},
+	"differentiation_medium":{
+	    "description": "Culture media used to induce a specific differentiation response.",
+	    "type": "string",
+	    "user_friendly": "Differentiation Media",
+	    "example": "RPMI 1640 + B27, Neurobasal Media, StemPro-34 Serum-Free Medium"
+	},
+	"differentiation_small_molecules":{
+	    "description": "Differentiation small molecules added to stem cell medium to induce a specific differentiation response.",
+	    "type": "string",
+	    "user_friendly": "Small Molecules",
+	    "example": "Retinoic Acid, CHIR99021 (GSK-inhibitor), Activin A, BMP4"
+	},
+	"differentiation_reagents":{
+	    "description": "Differentiation reagents added to stem cell culture to induce a specific differentiation response.",
+	    "type": "string",
+	    "user_friendly": "Differentiation Reagents",
+	    "example": "Ascl1 (mouse), nuclear receptor subfamily 4 group A member 2(Nurr1), Lmx1a(mouse)"
+	},
+	"differentiation_cell_culture_id":{
+	    "description": "Unique identifier of the differentiated cell culture after differentiation reagents and methods are applied. ",
+	    "type": "string",
+	    "user_friendly": "Differentiated Cell Culture ID"
+	},
+	"differentiation_target_pathway":{
+	    "description": "Targeted pathway for specific differentiation response.",
+	    "type": "string",
+	    "user_friendly": "Target Pathway",
+	    "example": "Wnt Pathway"
+	},
+	"differentiation_validation_method":{
+	    "description": "Method used to validate origin cell successfully differentiated to target cell.",
+	    "type": "string",
+	    "user_friendly": "Differentiation Validation Method",
+	    "example": "Pancreatic Cell DTZ Detection Assay, qPCR, Flow Cytometry, Immunocytochemistry Staining"
+	},
+	"differentiation_validation_results":{
+            "description": "Results confirming successful differentiation to target cell type.",
+            "type": "string",
+            "user_friendly": "Validation Results",
+            "example": "CD103 Positive, Nestin Positive, HCN4 Positive, CD11C Negative"
+        },
+	"differentiation_extracellular_substrate_matrix":{
+	    "description": "Extracellular substrate matrix in which cells are grown during differentiation.",
+	    "type": "string",
+	    "user_friendly": "Extracellular Substrate Matrix",
+	    "example": "Matrigel, Laminin, Vitronectin"
+	}
+    }
+}
--- a/json_schema/type/process/biomaterial_collection/differentiation_process.json
+++ b/json_schema/type/process/biomaterial_collection/differentiation_process.json
@@ -3,12 +3,12 @@
     "description": "Contains information on how a pluripotent cell is differentiated to a desired cell type or organoid.",
     "additionalProperties": false,
     "required": [
-		"describedBy",
-		"schema_type",
-		"process_core",
-		"differentiation_method",
-		"differentiation_cell_source_type",
-		"differentiation_target_cell_type"
+      "describedBy",
+      "schema_type",
+      "process_core",
+      "differentiation_method",
+      "differentiation_cell_source_type",
+      "differentiation_target_cell_type"
     ],
     "title": "differentiation_process",
     "type": "object",
--- a/json_schema/type/process/biomaterial_collection/differentiation_process.json
+++ b/json_schema/type/process/biomaterial_collection/differentiation_process.json
@@ -53,9 +53,12 @@
 	    "type": "string",
 	    "user_friendly": "Cell Source ID"
 	},
-	"differentiation_cell_source_type":{
-            "description": "Cell type of sample prior to differentiation to target cell.",
-            "type": "string",
+	"differentiation_source_cell_type":{
+            "description": "Cell type of sample prior to differentiation to target cell. Should be a child term of https://www.ebi.ac.uk/ols/ontologies/clo/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FCL_0000003.",
+            "type": "array",
+            "items": {
+		        "$ref": "https://schema.humancellatlas.org/module/ontology/5.1.0/cell_type_ontology"
+	        },
             "user_friendly": "Cell Source Type"
 	    "example": "hESC, iPSC, Mesoderm Progenitor, Neural Progenitor"
         },
--- a/json_schema/type/process/biomaterial_collection/differentiation_process.json
+++ b/json_schema/type/process/biomaterial_collection/differentiation_process.json
@@ -39,7 +39,7 @@
 		"differentiation_method": {
 			"description": "Differentiation method applied to stem cell culture to induce a specific differentiation response.",
 			"type": "string",
-			"user_friendly": "Differentiation Method",
+			"user_friendly": "Differentiation method",
 			"example": "Embryoid Body, Monolayer, Inductive Co-Culture"
 		},
 		"differentiation_cell_source_id":{
@@ -53,7 +53,7 @@
             "items": {
 		        "$ref": "https://schema.humancellatlas.org/module/ontology/5.1.0/cell_type_ontology"
 	        },
-            "user_friendly": "Cell Source Type",
+            "user_friendly": "Cell source type",
 			"example": "hESC, iPSC, Mesoderm Progenitor, Neural Progenitor"
         },
 		"differentiation_target_cell_type":{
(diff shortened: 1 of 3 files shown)


json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,137 @@
...
+            "type": "object",
+            "$ref": "https://schema.humancellatlas.org/core/process/5.1.0/process_core"
+        },
+	"induced_pluripotent_cell_induction_protocol_name":{
+	    "description": "Unique identifier of the protocol file which captures the details of how the cellular transformation was performed.",
+	    "type": "string",
+	    "user_friendly": "Induction Protocol Name",
+	    "example": "ThermoFisher Scientific- Generation of hiPSCs using the Epi5 Episomal iPSC Reprogramming Kit"
+	},
+	"induced_pluripotent_cell_source": {
```
daniwelter: Is this always going to be a cell type? If so, this should be explicitly stated in the description and the cell type ontology module used

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,137 @@
...
+		"piggyBac transposon",
+		"miRNA viral",
+		"adenovirus",
+		"cre-loxP",
+		"plasmid",
+		"retroviral"
+	    ],
+	    "user_friendly": "Induction Method"
+	},
+	"induced_pluripotent_cell_induction_kit_manufacturer":{
```
daniwelter: Consider using the purchased_reagents module for any appropriate fields here and using new fields only if they are not already in the module

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,137 @@
...
+	    "type": "string",
+	    "user_friendly": "Induction Reagents",
+	    "example": "POU5F1, SOX2, KLF4, c-MYC"
+	},
+	"induced_pluripotent_cell_culture_id":{
+	    "description": "Unique identifier of the induced pluripotent stem cell culture or cell line after induction reagents and methods are applied.",
+	    "type": "string",
+	    "user_friendly": "iPSC Culture ID"
+	},
+	"pluripotency_assay":{
```
daniwelter: We usually try to avoid the word "assay" as it is overloaded in the HCA context. Would pluripotency_test work?

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,137 @@
...
+	    "description": "Confirmation whether a viral vector has been removed after induction.",
+	    "type": "string",
+	    "enum": [
+		"yes",
+		"no",
+		"unknown"
+		],
+	    "user_friendly": "Pluripotent Vector Removed"
+	}
+	"induced_pluripotent_cell_induction_produced_in_house":{
```
daniwelter: why is this fields relevant?

json_schema/type/process/biomaterial_collection/differentiation_process.json
```diff
@@ -0,0 +1,134 @@
...
+	    "example": "Abcam- Human embryonic stem cells to hepatic endoderm protocol"
+	},
+	"differentiation_method": {
+	    "decription": "Differentiation method applied to stem cell culture to induce a specific differentiation response.",
+	    "type": "string",
+	    "usert_friendly": "Differentiation Method",
+	    "example": "Embryoid Body, Monolayer, Inductive Co-Culture"
+	},
+	"differentiation_cell_source_id":{
+	    "description": "Unique identifier of the cell source prior to differentiation to target cell.",
```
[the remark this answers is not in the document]
pnejad: User-supplied ID!

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
pnejad: Test or screen could work!

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
pnejad: I added this based on feedback I received! Please see the document I shared with you during our call.

malloryfreeberg:
@willrockout @pnejad Thanks for adding a new timecourse module! One thing that's missing is adding a reference to the timecourse module in all the biomaterial type schemas. Can you of you do this update?

willrockout:
@malloryfreeberg I was wondering about this it seems fairly repetitive to have time course in every biomaterial type. I think it would work to just have it in the bottom types so cell_suspension or cell_line. What do you think?

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
willrockout: @pnejad can you weigh in on this

willrockout: @pnejad

json_schema/type/process/biomaterial_collection/differentiation_process.json
```diff
@@ -0,0 +1,134 @@
...
+	    "user_friendly": "Differentiation Media",
+	    "example": "RPMI 1640 + B27, Neurobasal Media, StemPro-34 Serum-Free Medium"
+	},
+	"differentiation_small_molecules":{
+	    "description": "Differentiation small molecules added to stem cell medium to induce a specific differentiation response.",
+	    "type": "string",
+	    "user_friendly": "Small Molecules",
+	    "example": "Retinoic Acid, CHIR99021 (GSK-inhibitor), Activin A, BMP4"
+	},
+	"differentiation_reagents":{
```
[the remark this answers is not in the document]
malloryfreeberg: @willrockout yes. There technically shouldn't be a separate "purchase_reagents" tab in the spreadsheet. Instead, a field referencing the purchased_reagents module should appear in the spreadsheet with the field name prepended to all the fields in the purchased_reagents module. At least, this is the plan :)

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,136 @@
...
+            "description": "Core process-level information.",
+            "type": "object",
+            "$ref": "https://schema.humancellatlas.org/core/process/5.1.0/process_core"
+        },
+	"induced_pluripotent_cell_induction_protocol_name":{
+	    "description": "Unique identifier of the protocol file which captures the details of how the cellular transformation was performed.",
+	    "type": "string",
+	    "user_friendly": "Induction Protocol Name",
+	    "example": "ThermoFisher Scientific- Generation of hiPSCs using the Epi5 Episomal iPSC Reprogramming Kit"
+	},
```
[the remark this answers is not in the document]
malloryfreeberg: What you can have here is a name for the induction process itself - not a link to a PDF - which is actually a field in the process_core schema (process_name). 

```
        "process_name": {
            "description": "A short, descriptive name for the process that need not be unique.",
            "type": "string",
            "user_friendly": "Process name"
        },
```

json_schema/type/process/biomaterial_collection/differentiation_process.json
```diff
@@ -0,0 +1,133 @@
...
+	"differentiation_protocol":{
+	    "description": "Unique identifier of the protocol file which captures the details of how differentiation of cells were performed.",
+	    "type": "string",
+	    "user_friendly": "Differentiation Protocol",
+	    "example": "Abcam- Human embryonic stem cells to hepatic endoderm protocol"
+	},
+	"differentiation_method": {
+	    "decription": "Differentiation method applied to stem cell culture to induce a specific differentiation response.",
+	    "type": "string",
+	    "usert_friendly": "Differentiation Method",
```
malloryfreeberg: `usert_friendly` misspelled.

malloryfreeberg:
@willrockout @pnejad please run the `schemas_are_valid_json.py` script in the repo (in src/) prior to making any PRs. There are two missing commas in the differentiation and induction schemas. I also found 2 instances of `usert_friendly` instead of `user_friendly` that need to be fixed. We currently don't have a check for this in the `schemas_are_valid_json.py` script, but we could add it soon (pinging @daniwelter ).

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,136 @@
...
+	    "description": "Induction reagents added to primary cell culture to induce pluripotency.",
+	    "type": "string",
+	    "user_friendly": "Induction Reagents",
+	    "example": "POU5F1, SOX2, KLF4, c-MYC"
+	},
+	"induced_pluripotent_cell_culture_id":{
+	    "description": "Unique identifier of the induced pluripotent stem cell culture or cell line after induction reagents and methods are applied.",
+	    "type": "string",
+	    "user_friendly": "iPSC Culture ID"
+	},
```
malloryfreeberg: @pnejad For `induced_pluripotent_cell_culture_id`, do you mean an ID of the cell culture/cell line produced by this iPSC induction process? i.e. the output biomaterial?

lauraclarke:
@pnejad @willrockout I wonder if it is worth taking a step back and you describing what important information you are trying to capture about the ipsc differentiation process outside of the consideration of what fields are needed 

At the moment it feels a little like we are digging into the detail without understanding the bigger picture

What information do you want to capture and who needs this info?

lauraclarke:
@pnejad a document which specified the requirements would be really helpful as I am missing context and I think everyone already has enough meetings, I don't want to retread already covered ground

lauraclarke:
Paris, I suspect the most useful thing might be for me is a version of this diagram for the ipsc differentiation process you are trying to model would help me alot

example metadata entity graph

json_schema/type/process/biomaterial_collection/differentiation_process.json
```diff
@@ -0,0 +1,133 @@
...
+            "description": "Core process-level information.",
+            "type": "object",
+            "$ref": "https://schema.humancellatlas.org/core/process/5.1.0/process_core"
+        },
+	"differentiation_protocol":{
+	    "description": "Unique identifier of the protocol file which captures the details of how differentiation of cells were performed.",
+	    "type": "string",
+	    "user_friendly": "Differentiation Protocol",
+	    "example": "Abcam- Human embryonic stem cells to hepatic endoderm protocol"
+	},
```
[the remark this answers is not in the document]
willrockout: @malloryfreeberg I want to double check that this field should be removed and nothing else needs to be added to the protocol module correct?

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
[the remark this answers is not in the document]
willrockout: @malloryfreeberg Same as the question above.

json_schema/type/process/biomaterial_collection/differentiation_process.json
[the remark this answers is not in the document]
malloryfreeberg: Correct :)

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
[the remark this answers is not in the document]
malloryfreeberg: Correct :) The iPSC induction process name will be encoded by `process_core.process_name`.

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
willrockout: @malloryfreeberg This ID is in reference to the output from this process which is important to capture as it is what you work with moving forward. This allows any cells that are taken from it and used to be traced back to that batch. After talking with @pnejad I think this is best placed in the cell_suspension, cell_line, and organoid biomaterial types. Let me know your thoughts?

lauraclarke: @willrockout 

If the sample hierachy is

donor (assuming it exists) -> ipsc -> differentiated cells -> cell suspension which tends to suggest the ipsc which was used to produce the cell suspension is already part of the hierarchy

Can you explain why the ipsc id you are proposing won't already be part of the hierarchy we record?

lauraclarke: Just realised this got quite long, the tldr bit is every ipsc cell line from a catalogue should have its own biomaterial record which points to it's catalogue record.

The sample hierarchy for the ipscs should be captured as biomaterials.
There should at least by ipsc line (e.g https://www.phe-culturecollections.org.uk/products/celllines/ipsc/detail.jsp?refId=77650042&collection=ecacc_ipsc) which outputs either a cell suspension directly or a differentiated cell culture which is used to create the cell suspension and if we have the information we should also capture the donor (e.g http://www.hipsci.org/lines/#/donors/HPSI-bezi)

There should never be a cell culture or cell suspension which appears without a source even if that source if a purchased vial of cell lines so the identity of the cell line used should be captured as part of the biomaterial which the cell suspension is derived from not as part of the process which is used to derive the cell suspension/cell culture

Does this make sense? For every ipsc there should be an ipsc biomaterial

lauraclarke: Just to add then a user can infer which ipsc was used to produce the differentiated cells which were assays using the sample hierachy

willrockout: Okay, I think where the confusion is coming is mapping each step in that hierarchy to a biomaterial. Kate=Donor but the ipsc/differnetiated cells would be the cell suspension correct? Meaning currently we would only capture the ID for the differentiated cell but miss the ID for the ipsc's. Because of at least to my understanding if we have a biomaterial that undergoes multiple processes we only capture the end product. Also, I'm in full agreement that this field should be moved into a biomaterial tab preferably cell_suspension, cell_line, and organoid.

willrockout:
@malloryfreeberg @daniwelter Can you check this again I believe I have fixed all the problems and made the necessary changes. Only open one is iPSC Culture ID which I believe we can remove and insert into biomaterials.

malloryfreeberg:
I took a stab at modeling an experiment using iPSCs with the HCA metadata entity model and showing where some fields from the processes/biomaterials might go. @pnejad can you confirm whether this fits with how you are envisioning iPSC-based samples will work? If it's wrong, can you let me know what needs to be changed? 
ipsc_model

malloryfreeberg:
@willrockout some notes on the differentiation process schema:

1. `differentiation_cell_source_id`: We don't need this field here because differentiation process information will go in the same tab as the cells being differentiated. Therefore, there is no need to have a linking field.

1. `differentiation_cell_culture_id`: If this field is meant to point to the output biomaterial, then we don't need this field in the process schema. This link will be determined based on the `has_input_biomaterial` field and by this process being in the same tab as the input biomaterial.

1. `differentiation_source_cell_type`: We don't need this field in the process schema. `cell_type` lives in the schema for the biomaterial (cell line or cell suspension) that is the input (i.e. source) to the differentiation process.

1. `differentiation_target_cell_type`: We should discuss whether to keep this field in this schema. `cell_type` is currently a field on biomaterial (cell line or cell suspension) which would be the output from the differentiation process. We have to decide whether it will be too confusing to have a `target_cell_type` field on the process in addition to a `cell_type` field on the biomaterial. Is it possible these two fields will ever not be the same?

In general, process/protocol schemas will never have fields that are links to biomaterials. All linking will be defined by spreadsheet fields *only* and then translated into a single JSON document following the `links.json` schema.

json_schema/type/process/biomaterial_collection/differentiation_process.json
```diff
@@ -0,0 +1,134 @@
...
+	    "user_friendly": "Differentiation Media",
+	    "example": "RPMI 1640 + B27, Neurobasal Media, StemPro-34 Serum-Free Medium"
+	},
+	"differentiation_small_molecules":{
+	    "description": "Differentiation small molecules added to stem cell medium to induce a specific differentiation response.",
+	    "type": "string",
+	    "user_friendly": "Small Molecules",
+	    "example": "Retinoic Acid, CHIR99021 (GSK-inhibitor), Activin A, BMP4"
+	},
+        "process_reagents": {
```
malloryfreeberg: Please rename `process_reagents` to `differentiation_reagents`. This is in keeping with specific naming of reagents, like in the library_preparation_process schema, e.g.:

```
        "library_construction_kit": {
            "description": "Kit used to construct the sequencing library.",
            "type": "object",
            "$ref": "https://schema.humancellatlas.org/module/process/5.1.0/purchased_reagents",
            "user_friendly": "Library construction kit"
        },
```

willrockout: Interesting I had copied that from another process dissociation I believe that's why I had left the name

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,134 @@
...
+	    "description": "Confirmation whether a viral vector has been removed after induction.",
+	    "type": "string",
+	    "enum": [
+		"yes",
+		"no",
+		"unknown"
+		],
+	    "user_friendly": "Pluripotent Vector Removed"
+	},
+	"induced_pluripotent_cell_induction_produced_in_house":{
```
malloryfreeberg: If this field can only be "yes" or "no", it should be a boolean type. See the `is_living` field in the `donor_organism` schema for the syntax.

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,134 @@
...
+	    "type": "string",
+	    "user_friendly": "iPSC Culture ID"
+	},
+	"pluripotency_test":{
+	    "description": "Description of how pluripotency was tested in induced pluripotent stem cells.",
+	    "type": "string",
+	    "user_friendly": "Pluripotency Test",
+	    "example": "teratoma assay"
+	},
+	"pluripotency_assay_results":{
```
malloryfreeberg: Since `pluripotency_assay` was renamed to `pluripotency_test`, can you change `pluripotency_assay_results` to `pluripotency_test_results`?

json_schema/type/process/biomaterial_collection/differentiation_process.json
malloryfreeberg: Hmm. ok. We'll need to rename those then :)

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
malloryfreeberg: @willrockout we definitely need to capture both the input cells (e.g. fibroblasts), the induced pluripotent cell line (iPSC), and the differentiated cells all as unique biomaterials. We then capture metadata for the pluripotency induction and differentiation processes, respectively, between those biomaterials. The metadata entity model is flexible, so we can capture as many intermediate biomaterials as we or the submitters need.

malloryfreeberg:
@willrockout some notes on the induced pluripotency process schema:

1. `induced_pluripotent_cell_source`: As for `differentiation_source_cell_type` in the differentiation process schema, we don't need this field in the process schema. `cell_type` will live on the biomaterial (cell line or cell suspension) that is the input (i.e. source) to the induction process.

1. `induced_pluripotent_cell_induction_kit_manufacturer`, `induced_pluripotent_cell_induction_kit`, `induced_pluripotent_cell_induction_kit_catalog_number`, `induced_pluripotent_cell_induction_kit_lot_number`, `induced_pluripotent_cell_induction_kit_titer`, and `process_reagents`: All of these fields should be removed and replaced with a single `induced_pluripotent_cell_induction_kit` field like you did for the differentiation process schema. The `purchased_reagents` module doesn't have a `titer` field, so feel free to add that field to the `purchased_reagents` module (this would be a minor update if the field isn't required).

1. `induced_pluripotent_cell_culture_id`: As for `differentiation_cell_culture_id`, if this field is meant to point to the output biomaterial, then we don't need this field in the process schema. This link will be determined based on the `has_input_biomaterial` field and by this process being in the same tab as the input biomaterial. We definitely need this field, it just doesn't belong in the process schema.

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
lauraclarke: So the original example had

Kate -> blood -> monocytes -> scRNA-seq fastqs

Where the arrows are processes

So the equivalent ipsc process would be

Kate -> skin -> fibroblast -> ipsc colony -> ipsc cell suspenion -> scRNA-seq
or using the same biomaterials if the same donor/ipsc colony was used

Kate -> skin -> fibroblast -> ipsc colony -> cultured cardiomyctes -> scRNA-seq

The first 4 biomaterials are the same and would use the same process/protocols in both graphs. The last two would be different

malloryfreeberg:
@willrockout @pnejad @cjvillar I have updated the `differentiation_process` schema. The last thing I want to check is whether any of the following fields are redundant, or could be set to be enums:

```
		"differentiation_growth_environment":{
			"description": "Growth environment in which cells are differentiated.",
			"type": "string",
			"user_friendly": "Growth environment",
			"example": "scaffold, monolayer, suspension, hanging drops, collagen, engraftment"
		},
```
```
		"differentiation_growth_factors":{
			"description": "Growth factor added to stem cell culture to induce a specific differentiation response.",
			"type": "string",
			"user_friendly": "Differentiation growth factors",
			"example": "FLT3 Ligand CTS Recombinant Human Protein"
		},
```
```
		"differentiation_medium":{
			"description": "Culture media used to induce a specific differentiation response.",
			"type": "string",
			"user_friendly": "Differentiation media",
			"example": "RPMI 1640 + B27, Neurobasal Media, StemPro-34 Serum-Free Medium"
		},
```
```
		"differentiation_small_molecules":{
			"description": "Differentiation small molecules added to stem cell medium to induce a specific differentiation response.",
			"type": "string",
			"user_friendly": "Small molecules",
			"example": "Retinoic Acid, CHIR99021 (GSK-inhibitor), Activin A, BMP4"
		},
```
```
		"differentiation_extracellular_substrate_matrix":{
			"description": "Extracellular substrate matrix in which cells are grown during differentiation.",
			"type": "string",
			"user_friendly": "Extracellular substrate matrix",
			"example": "Matrigel, Laminin, Vitronectin"
		}
```

Also, can you please confirm that none of the above fields overlap or could be supported by fields in the `growth_conditions` [module](https://github.com/HumanCellAtlas/metadata-schema/blob/develop/json_schema/module/biomaterial/growth_conditions.json) (pasted below).

```
        "passage_number": {
            "description": "The number of passages the cell line has been through.",
            "maximum": 1000,
            "minimum": 0,
            "type": "integer",
            "user_friendly": "Passage number"
        },
        "growth_medium": {
            "description": "The solid, liquid, or semi-solid medium used to support the growth of the biomaterial.",
            "type": "string",
            "example": "lysogeny broth (LB) medium",
            "user_friendly": "Growth medium"
        },
        "mycoplasma_testing_method": {
            "description": "The method used for detecting mycoplasma contamination of a biomaterial culture.",
            "type": "string",
            "example": "Indirect DNA stain using Hoechst 33258 with 3T3 indicator cells",
            "user_friendly": "Mycoplasma testing method"
        },
        "mycoplasma_testing_results": {
            "description": "Results of mycoplasma testing of a biomaterial culture.",
            "type": "string",
            "example": "No spots of bright blue stain observed at high magnification",
            "user_friendly": "Mycoplasma testing results"
        },
        "drug_treatment": {
            "description": "Description of drugs added to the growth medium for this biomaterial.",
            "type": "string",
            "example": "100 ug/mL ampicillin",
            "user_friendly": "Drug treatment"
        }
```

malloryfreeberg:
@pnejad @willrockout @cjvillar I have made the final round of updates to all the modules in this PR. Please address my questions in the previous comment about redundant/enum fields in the differentiation_process schema. Once addressed, we can propose these new schemas on the appropriate Slack channels, wait for the review period to elapse, and then merge into `develop`.

willrockout:
@malloryfreeberg I've removed the growth-related tags that you mentioned above with the goal to move them into the growth conditions module. This will be done in another pull request when I get back next week. @pnejad want to add one more thing to this pull request clearance_of_reprogramming_factors if you can discuss that on tomorrows call then I believe this PR can go to final review.

json_schema/type/process/biomaterial_collection/ipsc_induction_process.json
```diff
@@ -0,0 +1,105 @@
...
+                "$ref": "module/process/5.1.0/purchased_reagents"
+            },
+            "user_friendly": "Additional reagents"
+        },
+        "clearance_of_reprogramming_factors_method": {
+            "description": "Method used to clear or remove reprogramming factors after iPSC generation.",
+            "type": "string",
+            "example": "Removal of piggyBac transposon by a transposase",
+            "user_friendly": "Method to remove reprogramming factors"
+        }
```
malloryfreeberg: @pnejad Please confirm whether the clearance_of_reprogramming_factors_method field looks OK to you.

malloryfreeberg: Removing this field for now. Can add later once we've discussed with experts.

PR #276, opened 2018-05-15
Added kit titer field to purchased reagents module

willrockout:
Which schema would you like to be changed?

/json_schema/module/process/purchased_reagents.json

What new field(s) need to be added?

Added Kit titer field

Details for requesting new field(s) outlined below
Field name (required): kit_titer

Field description (required): Appropriate titer and volume recommendations found in each kit's Certificate of Analysis. 

Whether the field should be optional or required (required): Optional

An example of valid value for this field (optional): Titer: Specification is 3.0x10 7 − 9.0x107 Cell Infectious Units/mL

Why is the change needed?
This is needed for the induction process to help understand the virus and how it works in the kit.

### DIFF
--- a/json_schema/module/process/purchased_reagents.json
+++ b/json_schema/module/process/purchased_reagents.json
@@ -46,6 +46,12 @@
             "format": "date",
             "example": "2018-01-31",
             "user_friendly": "Expiry date"
-        }
+        },
+	"kit_titer": {
+	    "description": "Appropriate titer and volume recommendations found in each kit's Certificate of Analysis.",
+	    "type": "string",
+	    "example": "Titer: Specification is 3.0x10 7 ",
+	    "user_friendly": "Kit Titer"
+	    }
 	}
 }


pnejad:
@willrockout @malloryfreeberg 

I would suggest calling it 'induction_kit_titer'. I don't believe every kit has a titer number. I also think we need to fix the numbers in the example (should be ...x10^7).

I've atached an example.
[L2110082_A16517.pdf](https://github.com/HumanCellAtlas/metadata-schema/files/2005670/L2110082_A16517.pdf)

malloryfreeberg:
In order to make the field and module more reusable, I vote sticking with `kit_titer` instead of making it more specific to the induction process. The way module fields render in the UI/spreadsheet, they will have the inheriting type schema prepended to the module field. So, in the case on inheriting the purchased reagents module into the ipcs schema, the field would look something like `induced_pluripotent_cell_induction_kit.kit_titer`, so it would be very obvious that in this case the kit titer is related to induction.

malloryfreeberg:
@daniwelter When do you think we'll be ready to start testing out the new pre-release/release process? We don't have a good SOP for this yet, and testing the SOP with this minor update will be useful, I think.

willrockout:
@malloryfreeberg The versions.json file is that how we're keeping track of the semantic versioning now or is it just a file to help us organize during this time of flux?

malloryfreeberg:
@willrockout The `versions.json` file is essential for the release process moving forward. It is the file that will be used by the release script to enter the correct `id` and absolute `$ref` fields into the schemas. You can check out more info about it on @daniwelter 's branch [here](https://github.com/HumanCellAtlas/metadata-schema/blob/dw-HCA-634_MetadataReleaseProcessSpec/docs/Metadata_schema_release_process.md#specification). We hope to implement the release process soon, at which point the version tracking file *should* be updated automatically. For now, I updated the `versions.json` file manually.

PR #294, opened 2018-05-21
Cell viability updates. Fixes #275.

### LINKED ISSUE #275: Add fields that HipSci uses for cell lines
[HPSI0214i-wibj_2.CofA.20160227.pdf](https://github.com/HumanCellAtlas/metadata-schema/files/1993637/HPSI0214i-wibj_2.CofA.20160227.pdf)

Attached file is an example Certificate of Analysis for an iPSC cell line from HipSci - specifically [this](http://www.hipsci.org/lines/#/lines/HPSI0214i-wibj_2) line.

The CoA has some extra fields that we might consider capturing, in general, for other iPSC cell lines.
- **lot_number** (or **batch_number**; different from `catalog_number`, which we have)
- **cell_line_identity_test** and **cell_line_identity_test_result** (e.g. Fluidigm, pass)
- **cell_viability_result** - we have `cell_viability_method` and `percent_cell_viability`, but HipSci also has whether the method test passed/failed

There are a few more tests that HipSci lines undergo, but the ones listed above I think are probably the most useful.

Thoughts @pnejad ?

malloryfreeberg:
This PR adds a new field and updates a field in the `cell_morphology` module. The changes are mostly motivated by the need to better capture metadata about cell viability testing methods, specifically from purchased cell lines. This PR (partly) Fixes #275 .

### Changed
Changed name of cell_viability field to percent_cell_viability to accurately reflect the value of this field.
### Added
Added a new field - cell_viability_result - to capture whether the cell_viability_method assay passed or failed.

This update is a major change due to the changing of the cell_viability field name.


json_schema/module/biomaterial/cell_morphology.json
```diff
@@ -23,25 +23,41 @@
...
             "type": "object",
             "$ref": "module/ontology/length_unit_ontology.json",
+            "example": "micrometer",
             "user_friendly": "Cell size unit"
         },
-        "cell_viability": {
+        "percent_cell_viability": {
             "description": "Percent of cells determined to be viable.",
-            "type": "number",
+            "type": "string",
```
hannes-ucsc: What's the reasoning behind changing the type to string? Typing it as a string would affect range queries in ES in non-intuitive ways.

hannes-ucsc: I'd prefer making this a ratio, like 0.99 instead of 99%, but I am not a domain expert.

And the name could be `cell_viability_ratio`, but again just an idea.

malloryfreeberg: This was decided to be made a string to allow users to enter a range of values, for example "90-95". We couldn't find a way in default JSON to type the field as integer or number but also allow ranges to be valid.

malloryfreeberg: Your point is taken, though. Ideally we would get a single number for this, but sometimes cell viability is calculated multiple times on the same sample, and either an average or a range in reported.

hannes-ucsc: Maybe a purist would model this as a oneOf between just a number and an object with two numbers. But take this with a grain of salt, I know very little about the domain or JSON schema for that matter.

malloryfreeberg: Alternatively, we could enforce having only a single number (integer or number) in this field. As I'm thinking about downstream querying use cases, I imagine a search query to find samples where percent_cell_viability is >90 would be much easier if this field was only ever a single number.

malloryfreeberg: I have been convinced by your querying argument :) And have changed this back to a number.

Issue #225, opened 2018-03-06
Add two fields to project core for funder and grant title.


cjvillar:
Adding request information:

"Hello!
I'd like to request an update to the metadata schema. The changes I'd suggest are as follows:

1 - this would be a change to the core "project" schema
2- Don't think there is a field that currently exists. 
3 - It would be a new field that is free text and something along the lines of "Funder and specific grant number" could be broken up into two fields (e.g. 1) Funder 2) grant title). 

4 - This change is being requested to help HCA funders, track and understand specifically what data they have supported. Ideally, a funder would be able to browse or identify exactly which data they have supported and also the specific grant/program that the data came out of. It will be critical to rationalizing funding and also a metric that funders can track to evaluate impact/effectiveness of specific programs. 

All the best,

Jonah"

malloryfreeberg:
Thanks for the suggestion! Please confirm that the new field address your need:

Field name: Grant name
Description: The name of the grant funding the project.
Example: NSF Grant 0903629
Required: No

cjvillar:
Should we create two separate fields one for the name and the other for the grant ID?

lauraclarke:
@cjvillar is correct, we should collect grant title, grant id and funder name

At somepoint, we will then want to review funder names and maybe make it a controlled vocabulary

I also expect that some people will be running these projects using multiple funding streams so we might need these things to be an array rather than just one of each

As I said, there is a plan to give funders their own ids that work like ORCID ids but I think if we have three fields, adding a funder id wouldn't be that much additional work

jonahcool:
@lauraclarke @cjvillar agree that the 3 fields is probably best. May even be able to introduce some CV for certain funder fields to help each funder quickly identify work that they are supporting.

freeman-lab:
congrats on your first github comment @jonahcool 👍

malloryfreeberg:
@lauraclarke @jonahcool I'm on board with adding grant ID and funder name fields. My concern with adding a grant title field is that the value for this field might be the same as the project title. In fact, the example we have give for the project title field is "A title of a grant proposal or publication." While I'm aware that a project title and grant title might not always be the same, I wonder whether we are better served asking submitters to spend their time/energy filling in fields more useful for describing their data rather than how they were funded.

[...]

PR #297, opened 2018-05-21
Mg 225 adding grant info fields

hewgreen:
Fixes (#225) adds a module for project funders called funder.json with three optional fields. In light of this I also slightly altered the description of project_title not to include funder wording to encourage the use of these fields.

This is a minor change to project.json and a new schema (funder.json is v1.0.0)

### Changed
Added a pointer to new module with three optional fields to track project funders

### Added
A new module with three optional fields to track project funders


json_schema/core/project/project_core.json
```diff
@@ -29,7 +29,7 @@
         "project_title": {
             "description": "A summary of the project in a sentence.",
             "type": "string",
-            "example": "A title of a grant proposal or publication. Approximately 30 words.",
+            "example": "A title of a publication. Approximately 30 words.",
```
daniwelter: I would add the approximate word count to the description rather than the example, and give an actual example, eg one of the project titles from the preview dataset.

malloryfreeberg:
@hewgreen What is the status of this PR? Are we moving ahead with free text strings for all fields, or attempting to use enums at all?

PR #300, opened 2018-05-21
Bug fix - ontology module updates

daniwelter:
We recently switched from using the public Ontology Lookup Service (OLS) to our own internal OLS version, which only uses HCA-relevant ontologies. This has temporarily broken ontology validation as several of the ontologies used as a whole before are now slimmed into either HCAO or HCA-specific EFO. I have updated the relevant ontology modules to use the correct ontologies and/or root classes. This update is a bug fix and therefore a patch increment.

### DIFF
--- a/json_schema/module/ontology/biological_macromolecule_ontology.json
+++ b/json_schema/module/ontology/biological_macromolecule_ontology.json
@@ -27,7 +27,7 @@
             "description": "An ontology term identifier in the form prefix:accession",
             "type": "string",
             "graph_restriction":  {
-                "ontologies" : ["obo:chebi", "obo:efo", "obo:obi"],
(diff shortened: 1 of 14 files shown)


malloryfreeberg:
Why are the inheriting schemas updated? Now that we've moved to relative URLs, I thought we wouldn't have to update paths to inherited schemas and therefore inheriting schemas wouldn't have to increment versions?

malloryfreeberg:
Also - just out of curiosity - does the order of ontologies listed in the `ontologies` field matter?

daniwelter:
As it turns out, it wasn't the ordering that was the problem, but the fact that in a couple of modules, the ontologies list was a single string rather than multiple values. This is now also fixed but I thought it would be good to have the primary target ontology in first place.

PR #302, opened 2018-05-23
removing has_input_biomaterial from biomaterial_core.json

simonjupp:
`has_input_biomaterial` is a bit of a red herring in the JSON schema. It is basically these for the old spreadsheet importer to work out how to link an input biomaterial. In the new spreadsheet importer this is no longer required as you will have a dedicated column like "Derived from Donor id". This logic lives with the spreadsheet builder/importer, not in the JSON schema. 

The links between biomaterials are represented in ingest through the Ingest API and in bundles via the `links.json` file so `has_input_biomaterial` is redundant. 

This field is ignored by the spreadsheet importer, so having it in the schema is misleading. 

### Removed
Removed the `has_input_biomaterial` property as the purpose is missleading and no longer required by the spreadsheet importer

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,10 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
 
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/develop)
 
+### [core/biomaterial/biomaterial_core.json - v6.0.0] - 2018-05-23
+### Removed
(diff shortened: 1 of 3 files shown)


malloryfreeberg:
@simonjupp can you remove the `has_input_biomaterial` field from the `/schema_test_files/biomaterial/test_pass_specimen_0.json` test file? This is causing Travis CI to fail.

malloryfreeberg:
PS I'm very happy we are removing this field as it always bothered me that we had this one linking field and no others in the content 👍

simonjupp:
done, tests are now passing

PR #304, opened 2018-05-24
process renamed to protocol in many headers

hewgreen:
I have updated the column headers that contain dot notation. There were quite a few discrepancies in the protocol tabs. often process.core was used and often process_id etc needed renaming too. I made these changes to all Alegria to test these examples.


malloryfreeberg:
@aaclan-ebi can you merge this PR if it looks correct to you? Thanks!

PR #308, opened 2018-05-24
Enrichment module

daniwelter:
Added enrichment ontology module and updated enrichment_protocol.enrichment_method to use module instead of enum. I added the module to the versions file but didn't update the dependent version numbers, or add version numbers to the change log.


malloryfreeberg:
This update is now in pre-release. Feedback will be requested until Tuesday 29 May.

Issue #329, opened 2018-06-04
Add "feeder_layer_type" to growth_conditions schema


pnejad:
**For which schema is a change/update being suggested?**
I would like to request an update to the growth_conditions.json schema.

**What should the change/update be?**
I would like to add a new field to this schema called feeder_layer_type.

**What new field(s) need to be changed/added?**

* Field name: feeder_layer_type
* Field description: Type of feeder layer cells biomaterial was grown on.
* Required: No
* Example: "MEF Feeder Cells", "JK1 Feeder Cells", "SNL 76/7 Feeder Cells", "Human Marrow Stromal Cells"
* CV or enum: No

**Why is the change requested?**

This change is requested to capture the type of feeder layer cells biomaterial is grown on. The type of feeder layer used may have different effects on biomaterial growth/proliferation.

malloryfreeberg:
Combine with PR #327

PR #331, opened 2018-06-05
Added field feeder_layer_type to module growth_conditions

### LINKED ISSUE #327: Add "feeder_layer" field to iPSC schema

I would like to request an update the the `ipsc_induction_process.json` schema.

I would like to add a new field to this schema called `feeder_layer`.

* Field name: `feeder_layer`
* Field description: **TBD**
* Required: No
* Example: "feeder-free"
* CV or enum: Yes, an enum. Accepted values should be: "feeder-free" and "feeder-dependent"

**TBD**

daniwelter:
Also updated changelog. Fixes #327 and #329

### Added
Added a new field - `feeder_layer_type` - to capture the type of feeder layer cells biomaterial is grown on as this may have different effects on biomaterial growth/proliferation.

#### NB
Versions.json was not updated. Please update for a minor release type after pre-release, update version number in changelog and add any dependent version updates.


pnejad:
@daniwelter I will work on the list of enums for this field. We need to extend my list of examples from issue #329.

daniwelter:
Thanks @pnejad. We intend to release this on Friday, so anything you can give us by then would be great.

malloryfreeberg:
We now have a bunch of pre-released things in develop that need to be released on Monday 11 June. Please **do not** merge this PR until after develop has been merged to master on Monday!

malloryfreeberg:
@pnejad do these replace all the examples you listed in the original GH issue?

pnejad:
@malloryfreeberg  Yes they do. Since we are using enums for this field, it would be best to use cell types instead of cell line names (my examples in the original issue).

lauraclarke:
This feels like a longer list than is ideal for an enum. Is there a better way to do this?

malloryfreeberg:
We could just have it be a free-text string field. I don't know if this type of field is ontologizable (or if ontologizable is a word).

pnejad:
@malloryfreeberg 

Are we still combining two fields? From my understanding, we decided to combine feeder_layer and feeder_layer_type during yesterdays call. Labs can fill in this field as "feeder-free" or "feeder-dependent:__one of the above options__". Is this still the plan?

malloryfreeberg:
@pnejad yes. If you click "Files changed" at the top of the page, you can scroll down to see what I added to the module (highlighted in green). There you will see that "feeder-free" is the first value in the enum list.

lauraclarke:
Just to confirm, we aren't asking submitters to say feeder_dependent are we?

If we are going to use just one field I feel it might be better to say feeder free or the name of the feeder cell type and drop the feeder_dependent bit

daniwelter:
If it weren't for feeder-free, we could potentially use the cell type ontology module here

malloryfreeberg:
@pnejad @daniwelter can we finalize this change and then pre-release, so that we can use it with our pilot datasets?

malloryfreeberg:
Discussing with Dani, this field is currently best modelled with a long enum as only a small subset (~7) of the terms are in the HCAO ontology at the moment. The ontology would also have to be pretty broad to encompass the different species/cell lines, and it couldn't include the "Feeder-free" value. I'll merge the PR now and we can maybe come back to this field at a later date to discuss whether a bespoke ontology should be built for it.

PR #351, opened 2018-06-07
Add licensing information to project 

daniwelter:
### Added
Added three new field, `license_url`, `license_description` and `license_icon_url`.

**NB**: Do not release until **after Monday 11 June 2018.**


malloryfreeberg:
@lauraclarke Do these fields look ok to you? They seem important for the pilot so I'd like to merge them as soon as possible.

lauraclarke:
@caaespin @hannes-ucsc will need to use these fields to add licensing information to project pages. They might be better approving this than me

daniwelter:
I'd like to put this PR on hold while we think about this a bit more. The way we intend to capture licensing information makes project potentially not the best place for these fields. Current alternatives include:
1. licensing module imported into project
2. stand-alone license type (requires a new license core etc)

We also need to think about when licensing information is filled in. Should this be injected into the metadata at import time? Or at export time? @danielvaughan @rolando-ebi @rdgoite @aaclan-ebi @tburdett @simonjupp thoughts?

daniwelter:
I have updated the implementation of these fields to a module imported by project.

daniwelter:
In light of the new provenance_core information, this might be a good place for adding licensing information. Discuss feasibility with devs and ask for comments.

malloryfreeberg:
@lauraclarke is this a requirement for phase 1? It might be a good time to reassess where this information might go now that we have `provenance_core` and new bundle structure out.

lauraclarke:
I don't think this is a requirement for phase 1 but if there is time to get it in that would be good. Do we know who is responsible for the data browser displays now Carlos is leaving. It would be good to see what their plans are for displaying this info now and how useful it would be for them to have it in the metadata.

If they don't have time to pull it out of the metadata right now it might be better to leave to phase II

json_schema/module/project/licensing.json
```diff
@@ -0,0 +1,38 @@
...
+            "type": "string",
+            "pattern" : "^(http|https)://schema.(.*?)humancellatlas.org/module/project/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/license"
+        },
+        "schema_version": {
+            "description": "The version number of the schema in major.minor.patch format.",
+            "type": "string",
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+        "license_url": {
```
hannes-ucsc: I'd add a license short name from something like https://spdx.org/licenses/ so Orange has something to display in the project listing.

json_schema/module/project/licensing.json
```diff
@@ -0,0 +1,38 @@
...
+            "type": "string",
+            "description": "URL of the license under which this data is available",
+            "example": "https://creativecommons.org/licenses/by/4.0/",
+            "user_friendly": "License URL"
+        },
+        "license_description": {
```
hannes-ucsc: I'd consider a length limit for this field so user can't dump the entire GPL license into this field by mistake.

malloryfreeberg:
Currently the fields in this module are:
```
"license_description": "Description of the license under which the dataset is made available",
"license_url": "URL of the license under which the dataset is made available",
"license_icon_url": "URL of the icon associated with this license"
```

I like @hannes-ucsc 's idea of having a "short_name" or "identifier" as it's called at https://spdx.org/licenses/. I don't really understand the need for the "description", unless this is where the long-form name is supposed to go. If that's the case, I can picture the following being sufficient:

```
{
   "identifier": "GPL-3.0-only",
   "description": "GNU General Public License v3.0 only",
   "url": "https://www.gnu.org/licenses/gpl-3.0-standalone.html",
   "icon_url": "https://www.gnu.org/licenses/gpl-3.0-icon.html"
}
```

Where "description" could be replaced with "full_name" or similar.

json_schema/module/project/licensing.json
lauraclarke: It doesn't necessarily stop this being a good idea but it is worth noting we aren't going to let submitters define their own license but will have a standard license which might change but will be defined centrally.

malloryfreeberg:
@lauraclarke @tburdett, myself and @daniwelter are struggling with where these fields should go in the schema. We are unsure about a few things including: what is being licensed (each type entity? a whole project? each submission?)? Will licenses change? Will users ever actively choose a license?

tburdett:
@malloryfreeberg I don't think I have enough information to answer your questions.  Fundamentally I think supporting _per dataset_ licensing restrictions is a bad idea, it comes with significant implied downstream consequences for sharing and aggregation, and these can be a nightmare when combining across multiple datasets (and therefore licenses).

Have we captured the requirements for supporting multiple licenses? I suspect the reason for your questions is that this is not clearly understood, but if we have a doc somewhere, can you (Mallory or @hannes-ucsc maybe?) point me at it? I've also been out of the loop, but we mandated CC-BY for the preview release, what led us to change this decision?

daniwelter:
@tburdett I don't think we changed our mind on CC-BY for now. I had a discussion with Laura regarding this a while ago and the use case we're trying to address is that IF we ever change to a different license, any data ingested prior to the license change would be CC-BY and we'd need a way of flagging this to users. We weren't necessarily envisaging a scenario where the license would change at the drop of a hat or multiple licenses would be in use in parallel.

lauraclarke:
The reason for having per dataset license fields was so if the licensing policy changed we could make sure which datasets were licensed with which policy could be unambiguously identified after the change was made rather than asking anyone to make a judgement call as to how a dataset was licensed depending on date or some other piece of information.

The license was intended to be per project as another granularity creates greater issues from a change perspective. This does of course create some problems if the licensing policy changes while a project in the course of submitting data.

It currently isn't the intention to give the submitter the right to choose a license but by putting the field at the project level if there is a clear value to do that, it is possible in the future.

The other advantage of it being along side the project metadata is then it is transparent as to how a dataset is licenced even out of context of our data portal

One short term solution for now as submitters can't define it is just to have boiler plate somewhere that the data portal can reference and only add these fields when a new license is needed and we need to differentitate between old datasets and new datasets. That does though carry the burden of needed to update all old datasets with the original boilerplate.

daniwelter:
I'm still very uneasy about having any metadata object that isn't filled in by users. I would therefore propose moving license.json to `system`. It could still be imported into project, in the same way as provenance, or it could be a stand-alone json file per bundle, like links.json. Thoughts?

lauraclarke:
The display use case is to have this information displayed alongside a project description so the reuse rights are clear. If the data portal team can easily access this data from somewhere else that this works fine. If they can't, it will be more difficult

daniwelter:
I have pushed a new proposal to this PR with, as suggested above, license.json living under /system. Once implemented, a stand-alone license.json file will be added to every bundle going into the datastore. It will not be linked to any of the other metadata files. 
@hannes-ucsc are you happy with these changes?

malloryfreeberg:
@justincc I believe these fields will need to be hardcoded in by ingest.

malloryfreeberg:
@hannes-ucsc Could you re-review this PR :)

malloryfreeberg:
> I still think we should remove the icon_url field as the icon for the license can easily be injected into the data browser at display time.

Ok. Will do.

Issue #340, opened 2018-06-07
Add differentiation_media to differentiation_protocol.json


daniwelter:
AI from wrangler call 7/6/18

**For which schema is a change/update being suggested?**

differentiation_protocol.json

**What should the change/update be?**

*New field*: differentiation_media

**What new field(s) need to be changed/added?**

* Field name: differentiation_media
* Field description: TBD
* Required: no
* Example: TBD
* CV or enum: no

**Why is the change requested?**
TBD

PR #356, opened 2018-06-07
[WIP] Added differentiation_media and small_molecules to diff protocol. Fix…

### LINKED ISSUE #342: Add small_molecule field to differentiation protocol
AI from wrangler update call 7/6/18. 

Should this be a free-text or an ontology field? Example data?

willrockout:
…es #340 and Fixes #342


malloryfreeberg:
@willrockout please make the user-friendly terms Sentence case, not title case, to be consistent with the schemas. Also, please update the changelog.md file. See one of the current PRs or the [changelog.md](https://github.com/HumanCellAtlas/metadata-schema/blob/master/changelog.md) file for examples.

willrockout:
@malloryfreeberg can you check the changelog fix not 100% i did it right

changelog.md
```diff
@@ -7,6 +7,14 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
...
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/develop)
 
+### [type/protocol/biomaterial_collection/differentiation_protocol.json - v?.+1.?] - 2018-06-11
+### Added
+Added small_molecules field.
+
+### [type/protocol/biomaterial_collection/differentiation_protocol.json - v?.+1.?] - 2018-06-11
+### Added
+Added differentiation_media field.
+
```
malloryfreeberg: This looks _almost_ right :) Adding two new optional fields here I would group together into 1 update as they are related and they are in the same PR. So, I would take out the second entry here and change the description of the first entry to "Added `small_molecule` and `differentiation_media` optional fields." You have the version update syntax right as per some earlier examples in our changelog, but I think we want to move to having it be `?+1` instead of just `+1` in the position of the change. So here it would be `v?.?+1.?`.

PR #365, opened 2018-06-11
Created 10x specific module

willrockout:
### Added
Created 10x specific module
### Added
Added 10x module


json_schema/module/process/sequencing/10x.json
```diff
@@ -0,0 +1,61 @@
...
+	    "type": "number",
+	    "example": 25,
+	    "user_friendly": "PCR cycles"
+	},
+	"expected_cell_count": {
+	    "description": "The number of cells expected to be observed in the data matrix",
+	    "type": "number",
+	    "example": 3000,
+	    "user_friendly": "Expected cell count"
+	}
```
malloryfreeberg: I think this `expected_cell_count` field is a duplicate of the `type/biomaterial/cell_suspension/total_estimated_cells` field. How are these fields different and why do we need one specifically on the 10x module?

json_schema/module/process/sequencing/10x.json
```diff
@@ -0,0 +1,61 @@
...
+	    "description": "Result of visual inspection of emulsion after 10x run.",
+	    "type": "string",
+	    "user_friendly": "Drop uniformity"
+	},
+	"pcr_cycles": {
+	    "description": "Number of pcr cycles used for amplification.",
+	    "type": "number",
+	    "example": 25,
+	    "user_friendly": "PCR cycles"
+	},
```
malloryfreeberg: PCR cycle count is arguably important for all assays, not just 10x. We have two PCR-related fields in the `type/protocol/sequencing/library_preparation_protocol` schema: `library_preamplification_method` and `cdna_library_amplification_method`. PCR is a valid value for both of these fields. We should consider having a PCR cycle field at this level and not in the 10x-specific module, unless there is a specific reason that 10x needs this field and no other techs do.

json_schema/module/process/sequencing/10x.json
```diff
@@ -0,0 +1,61 @@
...
+	"pooled_channels": {
+	    "description": "The number of channes pooled within a sequencing lane.",
+	    "type": "number",
+	    "example": 4,
+	    "user_friendly": "Pooled channels"
+	},
+	"drop_uniformity": {
+	    "description": "Result of visual inspection of emulsion after 10x run.",
+	    "type": "string",
+	    "user_friendly": "Drop uniformity"
```
malloryfreeberg: Please add an example for `drop_uniformity`.

json_schema/module/process/sequencing/10x.json
```diff
@@ -0,0 +1,61 @@
...
+            "description": "The version number of the schema in major.minor.patch format.",
+            "type": "string",
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+	"loading_concentration": {
+	    "description": "Concentration of cells loaded onto sequencer.",
+	    "type": "string",
+	    "example": "10 pM",
+	    "user_friendly": "Loading concentration"
```
malloryfreeberg: What is a "sequencer"? One flowcell? One lane on a flowcell? Does this assume that people will be doing Illumina sequencing and not some other technology like PacBio or nanopore?

json_schema/module/process/sequencing/10x.json
willrockout: @malloryfreeberg The reason for this field is that it is useful for downstream analysis. After thinking about it your right and I will remove it since we already capture it.

lauraclarke:
@willrockout If they use git I would encourage them to give their feedback directly here as it means there is no room for crossed wired due to multiple rounds of communication

json_schema/module/process/sequencing/10x.json
willrockout: @ambrosejcarr would you be able to supply an example of what you'd like to see for this field?

json_schema/module/process/sequencing/10x.json
```diff
@@ -0,0 +1,50 @@
...
+            "pattern" : "^(http|https)://schema.(.*?)humancellatlas.org/module/process/sequencing/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/smartseq2"
+        },
+        "schema_version": {
+            "description": "The version number of the schema in major.minor.patch format.",
+            "type": "string",
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+	"loading_concentration": {
+	    "description": "Concentration of cells loaded onto a single lane on a flowcell.",
```
malloryfreeberg: @willrockout Just want to confirm this fields is meant to capture concentration of cell in a "sequencing lane on a flowcell" and not a "channel on a 10x chip".

willrockout: @malloryfreeberg yes that is correct

willrockout: Sorry didn't see this comment it was hidden for some reason

malloryfreeberg: In that case, this field is not specific to 10X (non-10X experiments can also be sequenced on flowcells), and I would suggest moving it to the sequencing_protocol schema (in a different PR).

willrockout: Alright I'll remove it from this one then it should be ready to go

willrockout:
@malloryfreeberg Ready to go

PR #366, opened 2018-06-12
Updated readme docs for spreadsheet examples

### LINKED ISSUE #364: ReadMe's for current spreadsheet directories

/example/spreadsheets/current/template and filled

Add readme files to both directories with descriptions of the current files. 

File names with a description about each file.

Eases a users ability to grab the correct template/example spreadsheet for there needs.

malloryfreeberg:
This PR adds README markdown files to the repos containing example and template metadata spreadsheets. It fixes #364.


hewgreen:
This needs further work as it changes because if this is mined as the source of truth for integration tests there are other factors we need to consider. I'll be meeting with Dan to discuss this. Regarding building a full YAML, this needs to be automated. I may try to do this depending on the outcome of meeting with Simon and Dan. More work needed but merging this for now.

PR #391, opened 2018-06-29
Archive in place and all sheets passing validation in dev

hewgreen:
I have updated and tested 1 cell dummy test, 100 cell glio and 1000 cell glio in http://ui.ingest.dev.data.humancellatlas.org. All sheets pass validation.

I have removed sheets for integration tests from the 'examples' directory

I have added sheets currently used DCP wide from the DCP repo to the archive folder and time stamped.

I have updated and expanded all related readme's


hewgreen:
'filled' directory shouldn't exist here. This path should just be /metadata-schema/infrastructure_testing_files/current

I'm not sure why this is showing up as a conflict.

PR #395, opened 2018-07-03
Bundle restructure

daniwelter:
This PR contains the proposed bundle restructure, with the removal of the bundle metadata schemas and the insertion of a provenance_core block into each type entity. Please share any comments, thoughts or concerns below.


rolando-ebi:
@rdgoite @aaclan-ebi we need to test spreadsheet generation/conversion with these schemas

docs/jsonBrowser/bundle.md
```diff
@@ -61,7 +61,7 @@ references | An array of reference files. | array | no |  |  |  |
 ## ingest
 _Information added or generated at time of ingest._
 
-Location: bundle/ingest_audit.json
+Location: bundle/provenance_core.json
```
rolando-ebi: bundle -> core ?

json_schema/core/provenance_core.json
```diff
@@ -1,26 +1,26 @@
 {
     "$schema": "http://json-schema.org/draft-04/schema#",
-    "description": "Information added or generated at time of ingest." ,
+    "description": "Provenance information added or generated at time of ingest." ,
     "additionalProperties": false,
     "required": [
         "document_id",
-        "submissionDate"
+        "submission_date"
```
rolando-ebi: I think we need a "uuid" string here as well

rolando-ebi: Not required, but as one of the properties

docs/jsonBrowser/bundle.md
rolando-ebi: I mean should `bundle/provenance_core.json` be `core/provenance_core.json` ?

daniwelter: Ignore docs, they haven't been rebuilt yet. I can confirm that provenance_core definitely now lives in core.

hannes-ucsc:
I have a procedural question. Does it make sense for me to look at this before we finish the RFC process?

dosumis:
Would be good to see some documentation of how key queries will work with the new bundle structure.

dshiga:
Thanks for tagging me @daniwelter. As far as I can tell, this does not really impact the secondary analysis service, so it looks okay to me. I believe the provenance metadata is filled in by ingest for us.

malloryfreeberg:
Everything looks good to me. Adding in the PR that we decided to move the `links.json` and the `provenance.json` schemas to a new directory called `system/` to differentiate them from user-facing metadata schemas. The advantage is a more organized schema structure and less confusion about what schemas data submitters and consumers should be looking at. 

This PR should be merged ASAP to coincide with releases in ingest/data store/data browser changes to accommodate the new bundle structure. Thanks @hannes-ucsc @simonjupp @aaclan-ebi @caaespin @dshiga !

PR #396, opened 2018-07-03
Analysis process-protocol restructure

daniwelter:
I (re)created analysis_process as there are quite a few fields that are pre-run process specific and would otherwise necessitate the creation of a protocol per run of the analysis pipeline. Requested by mint team via slack (https://humancellatlas.slack.com/archives/C49TW3RV5/p1530012078000198). Please register comments or concerns on this PR.


json_schema/type/protocol/analysis/analysis_protocol.json
```diff
@@ -6,68 +6,13 @@
...
-                },
-                "memory": {
-                    "type": "string"
-                },
-                "docker_image": {
-                    "type": "string"
-                }
-            }
-        },
         "parameter": {
```
samanehsan: The definition for "parameter" is no longer used in this schema, so it looks like it can be removed

daniwelter: Done

json_schema/type/process/analysis/analysis_process.json
```diff
@@ -0,0 +1,161 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "Information about the analysis process",
+    "additionalProperties": false,
+    "required": [
+        "describedBy",
+        "schema_type",
+        "process_core",
```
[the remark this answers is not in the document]
daniwelter: Well spotted, now added.

json_schema/type/protocol/analysis/analysis_protocol.json
```diff
@@ -117,61 +41,15 @@
             "type": "object",
             "$ref": "module/ontology/process_type_ontology.json",
```
samanehsan: I think this should be `module/ontology/protocol_type_ontology.json` now

json_schema/type/process/analysis/analysis_process.json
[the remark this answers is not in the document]
samanehsan: thanks!

json_schema/type/protocol/analysis/analysis_protocol.json
daniwelter: No, this is actually correct. It's a bit of an anomaly but we are no longer particularly supporting the protocol branch of the ontology apart from very generic terms. The terms we're interested in are all in the process branch, so we're referencing the process type module.

samanehsan:
When we submit analysis results, we can add the `analysis_process.json` to the submission envelope's `/processes` endpoint. But when would the `analysis_protocol.json` file get created? Would it belong in the input bundle, or would we also add it to the `/protocols` endpoint?

daniwelter:
@samanehsan Yes, you would have a standard analysis_protocol that you would submit with every analysis run for a given pipeline version and it would be submitted to the protocols endpoint.
Are you ready for us to merge this into pre-release?

json_schema/type/process/analysis/analysis_process.json
```diff
@@ -0,0 +1,168 @@
...
+        "process_type",
+        "timestamp_start_utc",
+        "timestamp_stop_utc",
+        "input_bundles",
+        "reference_bundle",
```
samanehsan: Did you mean to include the `reference_bundle` field here or in the `analysis_protocol.json` file? It's listed as a required field in both schemas.

TimothyTickle: Agreed, that is confusing. If I understand the relationship between the analysis_protocol (static and does not change between runs) and analysis_process (an instantiation of that protocol that has parametric changes recorded; things that could change with runs), I think reference bundle should stay in the analysis_process.json (I know we currently have one reference but we seem to be positioning for multiple references given we are wrangling more than human data and if this is to be portable to other groups, they may have multiple references which may or may not require changes to the pipelines).

json_schema/type/process/analysis/analysis_process.json
```diff
@@ -0,0 +1,168 @@
...
+            "description": "Core process-level information.",
+            "type": "object",
+            "$ref": "core/process/process_core.json"
+        },
+        "process_type": {
+            "description": "The type of process. Must be a child term of https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FOBI_0000272.",
+            "type": "object",
+            "$ref": "module/ontology/process_type_ontology.json",
+            "user_friendly": "Process type"
+        },
```
TimothyTickle: Both process_type and protocol_type point to the EBI protocol ontologies listing. If these are the same, maybe just have protocol_type (and drop process_type) and the process links to the protocol for that information. Presumably, every process of a protocol will be the same type.

json_schema/type/protocol/analysis/analysis_protocol.json
```diff
@@ -6,88 +6,12 @@
         "describedBy",
         "schema_type",
         "protocol_core",
```
TimothyTickle: In the protocol core, there are entries for protocols.io. Is it expected that the pipelines would have entries in protocols.io, we are currently using Dockstore as the place people can access workflows in a documented way. Dockstore seems to be more appropriate for workflows and I am not sure what additional value is gained when duplicating efforts with protocols.io.

json_schema/type/process/analysis/analysis_process.json
daniwelter: That was an oversight. Now fixed.

json_schema/type/protocol/analysis/analysis_protocol.json
daniwelter: protocols.io is the resource we're trying to push as the de facto protocols repo but another source is fine as well. We can always add an additional field to analysis_protocol for you to capture the Dockstore id/DOI/URL

json_schema/type/process/analysis/analysis_process.json
daniwelter: Our policy is that each type entity has an ontologised type field so I'd prefer to leave the process_type field on the analysis_process for the sake of consisteny but I would be happy to make it optional so you don't have to fill it in.

TimothyTickle: Just pointing out the repeated information. If this makes sense for policies on your end that is value enough. Concern addressed.

json_schema/type/protocol/analysis/analysis_protocol.json
TimothyTickle: If protocols.io are optional and a dockerstore attribute fulfills your need here then I am ok with that solution. Comment addressed.

json_schema/type/process/analysis/analysis_process.json
samanehsan: Thanks @daniwelter! The `reference_bundle` should also be listed under the properties below: 

```
"reference_bundle": {	
    "description": "Bundle containing the reference used in running the pipeline.",	
    "type": "string"	
}
```

samanehsan: Also @dshiga brought up the fact that we are using a placeholder id for this field since this field is required, but we are not using reference bundles yet. Would it make sense if this field was optional? Or do we expect that all pipelines will use reference bundles?

TimothyTickle: I would expect this to be moved to something we use every time and would be afraid to make it optional, that information is extremely valuable and should be captured.

json_schema/type/process/analysis/analysis_process.json
lauraclarke: @daniwelter why isn't the value for process difference to the value for protocol? that seems odd

json_schema/type/process/analysis/analysis_process.json
lauraclarke: I agree with @TimothyTickle that we can't make it optional. Eventually, we will extend the validation to make sure it is real and potentially make sure it makes sense at least from a species/tax id perspective.

It sounds like getting reference bundles ironed out is something which needs to potentially be a phase 2 priority. We need to make an estimate of how much work it is so we can figure out how quickly it can be implemented @morrisonnorman, this sounds like a task for you.

json_schema/type/process/analysis/analysis_process.json
daniwelter: @lauraclarke This is a quirk of the ontology design choices we made for EFO. The way we need to model protocols in the ontology world isn't anywhere like as useful as the way we model processes, so we decided to focus on the process hierarchy and not develop the protocol hierarchy. In the current metadata model, any two biomaterials/files are connected by one process, which is linked to one or more protocols. The idea is therefore that the process will be annotated with a more general term (if at all - we often just imply the process) while the protocols are annotated with the more detailed ontology term.

json_schema/type/process/analysis/analysis_process.json
samanehsan: Ok, I wasn't sure if pipelines could use references that are not in the data store, or if reference bundles would be the only option once they are implemented. If it's the latter, then I do agree it makes sense to make the field required.

rexwangcc:
@malloryfreeberg Thanks for the pointer, this is what i'm looking for :)

PR #402, opened 2018-07-05
Supplementary file schema fixes #313

### LINKED ISSUE #313: New supplemental_file.json schema
In order to attach supplemental files to a project/submission, we might need a new `supplemental_file.json` schema. It would import `file_core.json` and would potentially need additional field(s) to allow it to be linked to the biomaterial/protocol/project that it is associated with.

hewgreen:
### Release notes:

This PR adds the type supplementary_file.json. This is a first suggestion. Discussion around this issue highlights the requirement for very flexible schema that encourages submitters to add supplementary files.

file_core satisfies most requirements. In addition to this I propose three new fields 'supplementary_type' (required), 'source_method_name' and 'data_source'.

**supplementary_type** is a enum for one of "Supplementary analysis", "Supplementary data" and "Supplementary metadata". This broad categorisation along with the link to the biomaterial helps users to understand what the file contains. The data viewer could group files based on this field.

**source_method_name** this field is ontology linked and should contain the method or technique that was used to derive the supplementary data/analysis. this will allow us to investigate types supplementary information across multiple datasets to scope further ingest use cases.

**data_source** this is an optional field to link to a database record. For example this is useful to link to health records or data records.

These three fields could be removed if they are too restrictive. I held off adding links to method pdf document because a supplementary file about a supplementary file felt too complex. I also didn't provide a way to easily reference analysis pipelines that may have been used to generate the supplementary information.

### DIFF
--- a/json_schema/type/file/supplementary_file.json
+++ b/json_schema/type/file/supplementary_file.json
@@ -48,7 +48,7 @@
             "user_friendly": "Supplementary type"
         },
         "data_source": {
-            "description": "A link to an external database record.",
+            "description": "A link to an external database record or analysis pipeline.",
             "type": "string",
             "example": "https://www.coriell.org/0/Sections/Search/Sample_Detail.aspx?Ref=GM17372&PgId=166",
             "user_friendly": "Data source"
--- a/json_schema/type/file/supplementary_file.json
+++ b/json_schema/type/file/supplementary_file.json
@@ -7,7 +7,6 @@
         "schema_type",
         "file_core",
         "supplementary_type",
-        "source_method_name"
     ],
     "title": "supplementary_file",
     "type": "object",
@@ -47,10 +46,15 @@
             "user_friendly": "Supplementary type"
         },
         "source_method_name": {
-            "description": "The method or technique used to derive the supplementary file.",
+            "description": "The method, technique used to derive the supplementary file.",
             "type": "object",
             "$ref": "module/ontology/process_type_ontology.json",
             "user_friendly": "Source method name"
+        },
+        "data_source": {
+            "description": "A link to an external database record.",
+            "type": "string",
+            "user_friendly": "Data source"
         }
     }
 }
--- a/json_schema/type/file/supplementary_file.json
+++ b/json_schema/type/file/supplementary_file.json
@@ -6,7 +6,8 @@
         "describedBy",
         "schema_type",
         "file_core",
-        "supplementary_type"
+        "supplementary_type",
+        "source_method_name"
     ],
     "title": "supplementary_file",
     "type": "object",
@@ -14,7 +15,7 @@
         "describedBy": {
             "description": "The URL reference to the schema.",
             "type": "string",
-            "pattern": "^(http|https)://schema.(.*?)humancellatlas.org/type/file/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/sequence_file"
+            "pattern": "^(http|https)://schema.(.*?)humancellatlas.org/type/file/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/supplementary_file"
         },
         "schema_version": {
             "description": "The version number of the schema in major.minor.patch format.",
@@ -45,17 +46,11 @@
             ],
             "user_friendly": "Supplementary type"
         },
-        "protocol_core" : {
-            "description": "Core protocol-level information.",
-            "type": "object",
-            "$ref": "core/protocol/protocol_core.json"
-        },
-        "source_method": {
-            "description": "The method or technique used to derrive the supplementary file.",
+        "source_method_name": {
+            "description": "The method or technique used to derive the supplementary file.",
             "type": "object",
             "$ref": "module/ontology/process_type_ontology.json",
-            "user_friendly": "Source method"
-        },
-
+            "user_friendly": "Source method name"
+        }
     }
 }
--- a/json_schema/type/file/supplementary_file.json
+++ b/json_schema/type/file/supplementary_file.json
@@ -41,15 +41,9 @@
             "enum": [
                 "Supplementary analysis",
                 "Supplementary data",
-                "Supplementary metadata"
+                "Supplementary metadata",
             ],
             "user_friendly": "Supplementary type"
-        },
-        "source_method_name": {
-            "description": "The method, technique used to derive the supplementary file.",
-            "type": "object",
-            "$ref": "module/ontology/process_type_ontology.json",
-            "user_friendly": "Source method name"
         },
         "data_source": {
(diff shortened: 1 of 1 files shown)


hewgreen:
I'd appreciate some thoughts on this before we merge from @claymfischer @cjvillar @malloryfreeberg @willrockout @pnejad.

See https://github.com/HumanCellAtlas/metadata-schema/blob/mg-supplementary_files/json_schema/type/file/supplementary_file.json

malloryfreeberg:
@hewgreen can you reformat the three new suggested fields in the standard form that we require? The format is:

For each **new** field requested, please provide the following:

* Field name: A suggested name for the new field.
* Field description: A short, but clear, description for the new field.
* Required: Whether the field should be required in the schema (yes, no)
* Example: An example value that would satisfy this field.
* CV or enum: Whether this field should be governed by a CV or an ontology. If yes, what should the CV/ontology be?

pnejad:
@hewgreen Would **source_method_name** be a required field if someone submits/fills out **supplementary_type**? 

One of our contributors in interested in submitting imaging files of their plates, along with spreadsheets with information imaging statistics, cell count and quality of that call. 

In their case, would **supplementary_type** = spreadsheet (xlsx) and **source_method_name** a list of all the methods/techniques used to generate the information within the spreadsheet? And how could the spreadsheet be linked to several .tif files? 

This contributor has 6 chips in their experiment (5k wells each) and spreadsheets for each chip.

hewgreen:
@malloryfreeberg there you go

**Field name:** supplementary_type
**Field description**: One of either 'Supplementary analysis', 'Supplementary data', 'Supplementary metadata'.
**Required**: yes
**CV or enum:** ["Supplementary analysis", "Supplementary data", "Supplementary metadata"]

**Field name:** source_method_name
**Field description**: The method, technique used to derive the supplementary file.
**Required**: no
**CV or enum:** using process_type_ontology.json

**Field name:** data_source
**Field description**: A link to an external database record.
**Required**: no
**CV or enum:** no

lauraclarke:
@hewgreen I think @pnejad highlights the issue with the supplementary_type field. She has interpreted to mean a file format not one of the three cases you specify

It isn't clear to me what the downstream use case for the supplementary_type field is? how do you expect it to be used. These are files which will always require human inspection to decide how to use them. If they are bound for automated processing we will need a specific file type and associated metadata not this form.

hewgreen:
@pnejad I think the Greenleaf case will be helped by allowing specimen_from_organism.state_of_specimen.gross_image and specimen_from_organism.state_of_specimen.microscopic_image on cell suspension (aka plate in their case). I think there may be other uses for this too. They can add their extra metadata files with this supplementary.json too. But with this proposal we won't capture the links from supplementary metadata and supplementary data.

@lauraclarke I want to help the user manually interrogating supplementary files to navigate them as best we can. I think this helps a bit. It also helps gives a flag that could be used by the data portal for some slightly fancier icons etc. It will also allow us to investigate supplementary type popularity across the HCA. Maybe these uses are not worthwhile, the fallback is to just use the core fields and rely 100% on manual navigation.

lauraclarke:
@hewgreen can ingest support the additional files on biomaterial already?

I am not sure what your proposed tags actually help people do, I think a lot of people will be confused as to what is metadata, what is data and what is analysis. What do you call a facs plot? what about cytoff or a spreadsheet with a bunch of biochem measurements.

I understand your aim but I am not convinced this will reduce the amount of manual inspection that is needed to figure out what a file is

lauraclarke:
@hewgreen you still haven't explained how people know which of the three general labels "Supplementary analysis", "Supplementary data", "Supplementary metadata" 
they should use and how this actually reduces the amount of time people have to interact directly with the files to understand what they are

This feels like a field that people will pick the first from the list from rather than actually figure out what it such be.

The description also contains zero information about what the different types are and in what circumstances to use them. The types themselves are insufficiently descriptive for the description to be so brief

I do not understand the value this field adds to anyone's use of supplementary files. Each file has a free text description as to what each file is it feels like a much more useful thing.

hewgreen:
This seems clear to me, not sure if it is best to discuss in person or here? Maybe some other opinions would be helpful.

To address your first point, if you have uploaded raw data but want to provide additional analysis of that data then the type of file is Supplementary analysis. If you want to provide additional data not supported by the HCA metadata schema e.g. CyTOF, FACS then you select Supplementary data. If you have a file with extra metadata about a donor (e.g. medical history) or other biological entity then you select Supplementary metadata. Mixes of these things similar to supplementary files uploaded with publications would most likely be classified as either Supplementary analysis or Supplementary data. Either would be fine.

I'd agree that all enum fields are ripe for abuse, especially if validation is difficult. However, removal of library_preparation_protocol.cell_barcode.white_list_file, enrichment_protocol.protocol_core.document, collection_protocol.protocol_core.document and all other fields in the metadata that take a supplementary file name as their input necessitates this typing of supplementary files because we are now loosing the previously inherent typing. For example, we need to record that a specific txt file attached to an entity is the white list as opposed to something else.

Controlled typing over free text is better because it allows consumers to search for extra file types explicitly. If a project is heavily reliant on supplementary information, this becomes invaluable to navigate their data.

A real example. Consider I have a study with an smFISH image for each specimen. These are only a few colours with no hybridisation rounds but they are critical to show my transcript of interest acting as the publication discusses. This type of image is not transcriptomics, so is not supported, yet a central piece of evidence. A user/wrangler wants to add these files to each entity in the metadata model. They raise the issue with metadata committer and have a specific string added to the supplementary_type field enum, say something like "smFISH image". This differentiates these from the other supplementary files they have and provides a consistent field for a user to find the images in the study. Free text would be nicer here for intra project use cases but inter project queries require some conformity. This enum is good for our current use cases but may be relaxed to free text at a later date without breaking previously ingested data.

Another usage example. Consider a donor with 3 associated supplementary files. A "Supplementary metadata" file, "Supplementary analysis" file and a "Cell barcode whitelist file". As a data consumer, I can infer the whitelist is on the wrong entity, if I'm lacking metadata I know where to look, other wise I could ignore it and the "Supplementary analysis" file is a mystery that needs exploring by hand. Lets say that I then find the metadata file very useful. Now I can download all of that file type for the project for the other donors. I would know which file is of this type because the data portal would have the option to show these together for a given collection.

Another more general example would be to more easily help answer the question 'what data do HCA contributors have, that isn't formally support?'. Variations of this question will be asked by tertiary portal developers, funders and anyone with an eye on the future of the HCA's capabilities. Again free text may be enough to support this but an enum does so better while the process is wrangler lead pre logical validation. Pre submission logical validation could be as simple as asking 'do all of the protocols in the project have either a link to protocols.io or a supplementary pdf file that describes the protocol'.

malloryfreeberg:
Closing because we addressed this PR with https://github.com/HumanCellAtlas/metadata-schema/pull/423

PR #408, opened 2018-07-06
added project role attribute fixes #352

### LINKED ISSUE #352: Roles for contributors
Change for contact.json

New field addition Role and orcid id

role - we might want an ENUM here something like (project_member, external_curator, dcp_wrangler) We should consult on what the best labels might people as people might want to differentiate between bioinformatician, maybe free text is best in the first instance

orcid id - the orcid id of an individual

This is to help provide credit to everyone including those who prepared the datasets for submission and evetually link back to orcid records

hewgreen:
Added two new fields to json_schema/module/project/contact.json

Ran schemas_are_valid_json.py

**Added** project role attribute as a new optional field in the contacts json. This is an enum with the following options:

                "principal investigator",
                "external curator",
                "HCA curator",
                "technician",
                "computational scientist",
                "experimental scientist",
                "clinician",
                "pathologist",
                "other"

**Added** orchid id as a free text field


malloryfreeberg:
Also, please update the changelog :)

gabsie:
Hi, I suggest the following, please note the order:

user friendly name: Project Role
description: Role of the individual within this project.
example: Enter one of the following: "principal investigator", "experimental scientist", "computational scientist", "clinician", "pathologist", "technician", "external curator", "HCA curator", "other".

let me know if this works.

lauraclarke:
@gabsie I would suggest we use a different term to curator for HCA as what the wranglers are doing isn't really curation. Is there a reason not to use wrangler? An alternative would be HCA ingest support

lauraclarke:
We probably want co investigator too as some people who aren't the lead PI on a project might prefer it/

gabsie:
Hi, Laura, everyone

This can be whatever the role you decide upon or the team members have.
I think Data broker in this (human role) context is a good one, but if Data wrangler is what you have already agreed on for the team members, feel free to have that one.
I also want to avoid as much as we can the HCA abbreviation, as it might be confusing for people.
Is that possible? 
A rule for any type of form is also that Other - should allow you to enter whatever you want.
I would encourage that as well.

Gabs

hewgreen:
So is this ok?

user friendly name: Project Role
description: Role of the individual within this project.
example: Enter one of the following: "principal investigator", "co investigator", "experimental scientist", "computational scientist", "clinician", "pathologist", "technician", "external curator", "Human Cell Atlas wrangler", "other".

should we have 'external curator' or just 'curator'?

lauraclarke:
I think we need some way to indicate the curator wasn't part of the team which conducted the original experiment so external makes sense but happy to find a better term than external if there is one

gabsie:
Just "curator" is great

hewgreen:
If curation is done pre data submission is it classed as curation? I'm not sure. I suggested 'curator' because the external is nearly always implied. Don't forget Laura that this is just role. Other fields capture separate institutions if required.

lauraclarke:
So this particular role is aimed at a specific use case, when we import data from the Single Cell Expression Atlas.

We could make the role name very specific and call it EMBL-EBI Single Cell Expression Atlas Curator but that is a bit cumbersome

I am not sure institutional affiliation will cut it to clarify for two reasons. The first there will be people associated with different roles all from EMBL-EBI, say we have a study where there is someone from John Marioni's group who did the analysis work so is a computational biologist, someone from the Expression atlas who curated the paper so is a curator and one of you or Mallory who did the final push into the HCA. 

That means it will be very difficult for a person to infer what is going on purely from the institution

The second is . from the PI from a contributing lab reviews the list and sees a curator who they have no idea who they at least some of them may complain and ask who they are or even request them to be removed

As I said, there might be a different word which is better than external but we need some way of indicating this person wasn't a part of the core project team that doesn't rely on them having a particular view of the word curator

malloryfreeberg:
@hewgreen I will merge this as soon as you update the changelog :)

PR #418, opened 2018-07-12
Contacts: email required; institution optional

### LINKED ISSUE #411: Make email address optional and institution required in contact.json
I would like to propose the following changes top contact.json

email address should be changed from required to optional

Institution should be changed from optional to required

This change is proposed as @pnejad made the suggestion that people who we don't directly interact with may be annoyed if we publish their email address without asking and the EBI wranglers have struggled to get email addresses for everything

The idea for changing institution to required as it means we can give people looking at a project some information about where a person was based so if they want to find them it will be easier. 

@hewgreen made the sensible suggestion, that when our validation system can do conditional between field validation that we make email required for the person who is labelled as the corresponding contributor, for now we just have to keep an eye and try and ensure the wranglers get at least one email address

malloryfreeberg:
In `contact.json` schema:
- Changed `email` to be optional
- Changed `institution` to be required. 
- Updated appropriate automatic test files.
- Fixes #411.


malloryfreeberg:
@daniwelter can you merge this PR? I updated the update_log.csv file and it's ready to go with the new automatic process. I can do the pre-release announcement on Slack after it's merged.

PR #420, opened 2018-07-13
Adding a new file type for protocol files fixes #422

### LINKED ISSUE #422: New file type for protocols

a new type/file protocol_file.json

Add a new file type with nothing other than the core file attributes.

We have received protocol pdfs and docx files with protocols that need to be captured in metadata. We currently have attributes in the metadata to record the file names as strings but this is not a mechanism for upload. Files need to be entities in their own right as are sequencing files to support upload. As protocols are quite distinct from supplementary files I propose that this should be a separate file type (not the same as supplementary files).

hewgreen:
We require a mechanisms for linking protocol pdfs and word documents etc to the metadata. This will replace the current mechanism whereby protocols have file names stored as strings. These will be later removed.

This new file type only contains core file fields.


hewgreen:
Note I have already ran release_prepare.py on this script. The version number is 1.0.0

PR #423, opened 2018-07-13
New supplemental_file.json

### LINKED ISSUE #422: New file type for protocols

hewgreen:
Added a generic supplemental file type is required to allow us to support supplementary file and protocol uploads.

An extra field has been added:

**attribute name**: file_description
**description**: A short description of this files content. This could include information about how the file was generated.
**required**: no

fixes #422


malloryfreeberg:
Replaces PR https://github.com/HumanCellAtlas/metadata-schema/pull/402

PR #428, opened 2018-07-16
  added a 10X integration spreadsheet and updated readme

hewgreen:
* Added new sheet to `metadata-schema/infrastructure_testing_files/current` for 10x example. Adapted `dcp_integration_test_metadata_1_SS2_bundle.xlsx`

* Updated README.


lauraclarke:
@hewgreen just a quick check, why is a SS2 example being adapted to represent 10X that sounds wrong

hewgreen:
We need a 10x example bundle. The SS2 sheet being adapted is dummy data.

lauraclarke:
Just to confirm, the SS2 original was entirely fake data?

Is there a reason we aren't using real data at this point?

malloryfreeberg:
@lauraclarke if we are going to use real data, we need to find a published 10x dataset to use. I don't think we should be using real, unpublished data - even a small subset of it - as our repo's infrastructure testing data. The contents of this test spreadsheet doesn't matter so much as long as the relevant fields are changed to reflect a 10x experiment. 

@hewgreen I tried ingesting/exporting this spreadsheet with some fake I1, R1, and R2 data, and it made 3 separate bundles, 1 for each file. I had to add an ID in the `process.process_core.process_id` column in the Sequence file tab (same ID in each row) to get ingest to export 1 bundle with all 3 files. Please update the spreadsheet accordingly. Thanks!

hewgreen:
The 10X sheet now passes now so we should merge as is. I included the yaml to ease future updating. I have not yet uploaded this with the data because we are waiting for the files and pointers to be added see https://github.com/HumanCellAtlas/dcp/issues/107.

PR #435, opened 2018-07-23
New file per entity bundles links_json

simonjupp:
Example JSON files for the file per entity bundle structure


simonjupp:
ignore this, I'll send another one soon with all the files.

PR #451, opened 2018-07-26
Familial name changes. Fixes #380

### LINKED ISSUE #380: Rename field is_parent_of to child and other relationship fields
### Original request
1. <is_child_of>
**Which schema needs to be changed?**
Change to familial_relationship.json
**What field in that schema need to be changed?**
is_child_of field
**What should the new field name be?**
Change to parent
Shortening. 
1. <is_parent_of>
**Which schema needs to be changed?**
Change to familial_relationship.json
**What field in that schema need to be changed?**
is_parent_of field
**What should the new field name be?**
Change to child
Shortening.
1. <is_sibling_of>
**Which schema needs to be changed?**
Change to familial_relationship.json
**What field in that schema need to be changed?**
is_sibling_of field
**What should the new field name be?**
Change to sibling
Shortening

### Consequences: 

1. Loss of explicit relationship directionality
2. Could be interpreted as a “yes/no” question, ie “is the individual a child?”
3. Breaking change for all downstream components

### Consequences: 

1. Loss of explicit relationship directionality
2. Could be interpreted as a “yes/no” question, ie “is the individual a parent?”
3. Breaking change for all downstream components

### Consequences: 

1. Loss of explicit relationship directionality
2. Could be interpreted as a “yes/no” question, ie “is/has the individual a sibling?”
3. Breaking change for all downstream components

cjvillar:
Changed Child to Parent, Parent to Child, and _sibling_of to sibling.


malloryfreeberg:
I think we have to change how this PR is being done. As it stands, this PR will bump the major version of the `familial_relationship.json` schema by 3, but there actually won't be any changes in the first and second major version bumps. Either (1) each of these 3 field name changes need to be their own PR with a single line in the `update_log.csv` file, such that merging each sequential PR will be 1 major version change of 1 field. Or (2) all 3 field changes need to be made in 1 PR with a single major version change. The former (1) is more in line with the "1 change per version bump" philosophy that the HCA is trying to follow. Therefore, each of the 3 field name changes need to be their own PR. @daniwelter do you agree with this approach?

daniwelter:
I'm wondering if it should actually be completely the other way - as all three changes are essentially of the same type, they should not only be in the same PR but in one single change, ie one line the update log, something like "Changed a to b, c to d and e to f".

malloryfreeberg:
I guess I just see it as cleaner to have a single major change per PR/version bump. For documentation patches, I'm less inclined to have a single change per PR/bump. But for major versions, it seems it would be easier/cleaner to track these changes as 1 per PR/bump. But, I could be swayed either way.

Point is, the PR as it stands should be changed. Either broken into 3 PRs, or merge the 3 lines in the `update_log.csv` in this PR to reflect 1 major version bump. @cjvillar

cjvillar:
Fixed update_log, Please review @malloryfreeberg @daniwelter @willrockout

PR #509, opened 2018-08-20
New example analysis process

malloryfreeberg:
- Added new analysis process example JSON


malloryfreeberg:
@daniwelter I can't get this example analysis_process.json to pass the test suite. Can you take a look at it?

daniwelter:
This is once again the bug in the python validator where it can't validate deeper than one import level. analysis_process imports analysis_file, which imports file_core and that messes up the base pass. I've never been able to find a workaround for this problem.

malloryfreeberg:
Thanks Dani. I commented out the test in the CI test for now, but we need to re-address this when we expand the test suite of tools for the schema.

PR #512, opened 2018-08-21
Fixed bug causing ontology schema validation issue

daniwelter:
Patch update to 4 ontology schemas to set include_self to true, plus a casing issue fix in the ancestry ontology. This fix is needed as root terms are not returned in OLS searches.


malloryfreeberg:
As @willrockout is out today, and this is just an ontology fix, I'm going to merge this w/o a second reviewer.

PR #518, opened 2018-08-28
Changed "disease" field to "diseases". Fixes #515.

### LINKED ISSUE #515: "disease" field renamed to "diseases"

I would like to request an update to the `donor_organism.json` and `specimen_from_organism.json` schemas.

The field `disease` in both schemas should be changed to `diseases` as they are arrays (can accept 1 or more disease values).

        "disease": {
            "description": "Short description of known disease(s) of the organism. Enter 'normal' if no known disease.",
            "type": "array",
            "items": {
                "$ref": "module/ontology/disease_ontology.json"
	        },
            "user_friendly": "Known disease(s)",
            "example": "bone squamous cell carcinoma"
        },

No new fields. 

We would like to adhere to our style guide which indicates that fields that can take on more than 1 value should be plural terms.

zperova:
Changed "disease" field to "diseases" in the  ```donor_organism.json``` and ```specimen_from_organism.json``` schemas.

Fixes #515.

 Passed the ```src/schemas_are_valid_json.py``` and ```src/json_examples_validate_against_schema.py``` scripts.


dosumis:
Looks fine. It's already an array, so (as per ticket), this is more accurate.

zperova:
I have changed the update_log. Thank you!

PR #519, opened 2018-08-28
Removed differentiation from field names. Fixes #517.

### LINKED ISSUE #517: Remove "differentiation" from field names in differentiation_protocol schema

I would like to request an update to 8 fields in the `differentiation_protocol.json` schema.

`differentiation_method ` -> `method`
`differentiation_media ` -> `media`
`differentiation_small_molecules ` -> `small_molecule`
`differentiation_target_cell_yield ` -> `target_cell_yield `
`differentiation_reagents ` -> `reagents`
`differentiation_target_pathway ` -> `target_pathway`
`differentiation_validation_method ` -> `validation_method `
`differentiation_validation_results ` -> `validation_results`

No new fields.

Having "differentiation" at the start of the indicated field names (A) is redundant with the schema name and (B) does not improve clarity of the field names. The word should be removed.

zperova:
Removed "differentiation" from 8 field names in the ```differentiation_protocol.json``` schema.

Fixes #517.

Passed the ```src/schemas_are_valid_json.py``` and ```src/json_examples_validate_against_schema.py``` scripts.


json_schema/type/protocol/biomaterial_collection/differentiation_protocol.json
```diff
@@ -39,53 +39,53 @@
             "type": "object",
             "$ref": "core/protocol/protocol_core.json"
         },
-        "differentiation_method": {
+        "method": {
```
daniwelter: This field should remain as differentiation_method in order to be consistent with other protocol types. All biomaterial_collection protocols currently have a x_method field so we risk overloading the field name if we remove the qualifier from it.

json_schema/type/protocol/biomaterial_collection/differentiation_protocol.json
```diff
@@ -39,53 +39,53 @@
...
         },
-        "differentiation_target_cell_yield": {
+        "target_cell_yield": {
             "description": "Percent of target cells obtained after directed differentiation of origin cell.",
             "type": "number",
             "example": 95,
             "user_friendly": "Percent target cell yield"
         },
-        "differentiation_reagents": {
+        "reagents": {
```
daniwelter: This should be renamed to protocol_reagents to maintain consistency with other protocol types.

zperova:
I have changed the update_log to correspond to the changes made so far until consensus is reached on this issue.

json_schema/type/protocol/biomaterial_collection/differentiation_protocol.json
malloryfreeberg: I actually might want to move towards changing all `protocol_reagents` fields to just `reagents`. In the programmatic, ingest calling of the fields, we will see, e.g.:

`differentiation_protocol.reagents`

and in the ES query we will see:

`"files.differentiation_protocol_json.reagents.<fieldname>": "<value."`

and in the spreadsheet we can append the `title` of the schema to the reagents field to get:

"Differentiation protocol Reagents"

Meaning we don't actually have to have the `protocol_` prefix for all the terms.

json_schema/type/protocol/biomaterial_collection/differentiation_protocol.json
```diff
@@ -39,53 +39,53 @@
...
         },
-        "differentiation_validation_method": {
+        "validation_method": {
             "description": "Method used to validate origin cell successfully differentiated to target cell.",
             "type": "string",
             "user_friendly": "Differentiation validation method",
             "example": "Pancreatic Cell DTZ Detection Assay, qPCR, Flow Cytometry, Immunocytochemistry Staining"
         },
-        "differentiation_validation_results": {
+        "validation_results": {
```
malloryfreeberg: Since this field takes a single value, please make it singular (part of the new style guide).

json_schema/type/protocol/biomaterial_collection/differentiation_protocol.json
daniwelter: Works for me

Issue #543, opened 2018-09-17
Update each document in docs/ folder as per Zina's reviews


malloryfreeberg:
See Zina's reviews [here](https://docs.google.com/document/d/1NEgYFrgDb0mM9HAhLoNE6X1G5N0hBERNXpMsOgNvNWY/edit).

Ticket includes updates to the following docs:
- [x] evolution.md 
- [x] rationale.md
- [x] release_process.md
- [x] structure.md 
- [x] committers.md
- [x] contributing.md

Issue #487, opened 2018-08-08
Finalize documentation in the metadata-schema repo for phase 1


zperova:
As a consumer of the metadata-schema repo I would like to see finalized documentation in order to understand the metadata standard.

PR #545, opened 2018-09-17
Documentation review: rationale.md. Part of #487 and #543

zperova:
Documentation review: review of the rational.md

Addresses Issue#543: Update each document in docs/ folder as per Zina's reviews, which is a part of Epic: Finalize documentation in the metadata-schema repo for phase 1#487

This is documentation change only. No tests needed.


malloryfreeberg:
I think we need to expand more on the section about decoupling the schema from other software components. We don't need a ton of detail, but this is a very important design principle that needs to be highlighted. thoughts @lauraclarke @tburdett @simonjupp ?

PR #552, opened 2018-09-20
Fix: File names in public beta examples should adhere to spec

hannes-ucsc:
- removed the UUID from the file name and replaced with numeric suffix
- removed duplicated extension (`.json.json`)


malloryfreeberg:
Thanks for the fix, @hannes-ucsc !

PR #555, opened 2018-09-20
first pass at a test plan for the metadata schema

diekhans:
Initial proposal for metadata schema test plan, for review only, please do not merge.

### DIFF
--- a/docs/test_plan.md
+++ b/docs/test_plan.md
@@ -1,0 +1,64 @@
+# The Human Cell Atlas: Metadata Test Plan
+
+## Table of Contents
+- [Introduction](#introduction)
+- [Develop tests](#develop_tests)
+- [Integration tests](#integration_tests)
+- [Stress tests](#stress_tests)
+- [Ideas for improvement](#ideas_for_improvement)
+
+## Introduction
+
+This document is an overview of *proposed* test process for the metadata
+schema.  The goal is to provide a test system that ensures the schema is
+correct and follows the style, depreciation, and versioning guidelines.
+
+The tests will be deployed using a continuous integration (CI) system that
+validates changes when are tested when pushed to the repository.  By using a CI
+system, there are no requirements on the developers or unexpected software.
+
+Documentation extraction is done the CI process, making it a product of the
+primary source, rather than checking it into the tree.
+
+
+**What is in this document?**
+ - The test plan for the metadata schema definition
+ - Goals and high-level definition of the metadata schema tests
+ 
+**Who should be reading this document?**
+ - Schema developers
+
+**What *isn't* in this document?**
+ - Specific details of the implementation of the tests
+ - Testing of metadata validation
+ - Testing of metadata tools and libraries outside of this repository
+
+## Develop tests
+These tests will be run as part of a pull request to the development
+repository and ideally configured to run when a developer pushes to a feature
+branch in a repository.
+
+The following are the set of tests to run on the schema.  Some of these may be combined into a single program.
+
+1. Schema syntax checking. Validate that the schema is a valid JSON schema
+1. Schema style lint.  Validate that the schema conforms to the style guide.  Configured known exceptions to the style guide generate warnings rather than errors.
+1. Example validation.  All examples in the tree are validated against the schema and must reflect the current version for each schema module.
+1. Validate and generate documentation.  Formatted documentation is an artifact of the CI build.
+1. Version compatibility validation. This compares the current version of the schema with the previous version, based on semantic version number.  It validates that the semantic versioning and depreciation rules have been followed.
+
+
+## Integration tests
+
+A pull request to integration runs the [Develop tests](#develop_tests) and additional runs the following test:
+
+1. Validation of a subset of real data.  This validates a specified set of production metadata against both the specified version and the latest compatible version (same major number) of the referenced schema. The subset is select based on a breadth of data types and previously failed examples.
+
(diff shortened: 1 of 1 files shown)


docs/test_plan.md
```diff
@@ -0,0 +1,64 @@
...
+repository and ideally configured to run when a developer pushes to a feature
+branch in a repository.
+
+The following are the set of tests to run on the schema.  Some of these may be combined into a single program.
+
+1. Schema syntax checking. Validate that the schema is a valid JSON schema
+1. Schema style lint.  Validate that the schema conforms to the style guide.  Configured known exceptions to the style guide generate warnings rather than errors.
+1. Example validation.  All examples in the tree are validated against the schema and must reflect the current version for each schema module.
+1. Validate and generate documentation.  Formatted documentation is an artifact of the CI build.
+1. Version compatibility validation. This compares the current version of the schema with the previous version, based on semantic version number.  It validates that the semantic versioning and depreciation rules have been followed.
```
malloryfreeberg: Currently, the person who make the pull request for a schema update asserts the type of update (major, minor, patch) which then is parsed by the `src/release_prepare.py` script to generate the appropriate increment in either the major, minor, or patch version number. What is meant by semantic version validation here? What exactly needs to be validated?

docs/test_plan.md
```diff
@@ -0,0 +1,64 @@
...
+## Develop tests
+These tests will be run as part of a pull request to the development
+repository and ideally configured to run when a developer pushes to a feature
+branch in a repository.
+
+The following are the set of tests to run on the schema.  Some of these may be combined into a single program.
+
+1. Schema syntax checking. Validate that the schema is a valid JSON schema
+1. Schema style lint.  Validate that the schema conforms to the style guide.  Configured known exceptions to the style guide generate warnings rather than errors.
+1. Example validation.  All examples in the tree are validated against the schema and must reflect the current version for each schema module.
```
malloryfreeberg: Example JSON documents have been difficult to maintain in the past. They are often not updated in step with schemas, and only cover the minimum required fields in a small subset of schemas. We need to think about a strategy for what kind of examples we want and how best to maintain them.

docs/test_plan.md
```diff
@@ -0,0 +1,64 @@
...
+1. Validate and generate documentation.  Formatted documentation is an artifact of the CI build.
+1. Version compatibility validation. This compares the current version of the schema with the previous version, based on semantic version number.  It validates that the semantic versioning and depreciation rules have been followed.
+
+
+## Integration tests
+
+A pull request to integration runs the [Develop tests](#develop_tests) and additional runs the following test:
+
+1. Validation of a subset of real data.  This validates a specified set of production metadata against both the specified version and the latest compatible version (same major number) of the referenced schema. The subset is select based on a breadth of data types and previously failed examples.
```
malloryfreeberg: This sounds like it requires running some `hca dss` commands to download/import bundles from the Data Store. Is that what you had in mind?

docs/test_plan.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Stress tests
+
+The stress tests are designed to do an extensive validation of existing metadata analogous to the DCP software stress tests.  These are run on a fixed schedule rather than being triggered by repository events.  If there have been no schema definition changes since the last run, the test is skipped.
+
+1. Validation of the full set of real data.  This validates all current production metadata against both the specified version and the latest compatible version (same major number) of the referenced schema
```
malloryfreeberg: This sounds like a great idea! It also sounds like a beast to do! What are the consequences of running such a test? Do we migrate up "failing" metadata to be compatible with the latest version? Do we record somewhere (maybe in the metadata JSON itself) when a production metadata JSON document becomes "invalid" against current schema and record something like 
```
"schema_version": "6.1.1",
"valid_schema_versions": "6.0.0-7.6.3",
```
which indicates the document was generated against version 6.1.1, but is valid against earlier version 6.0.0 (and nothing before) up to version 7.6.3, at which point after (maybe, e.g., v8.0.0) there was a backwards compatible-breaking update that means the document is no longer valid against the current v8.0.0 schema.

docs/test_plan.md
diekhans: The goal is to have software to catch changes that should be a major upgrade, but are not flagged as such.   I think this is doable, but it is still fuzzy in my mind on how.

malloryfreeberg:
@diekhans Is this PR still relevant? Or has the content been replaced by the more developed [Integration testing for metadata schema changes RFC](https://github.com/HumanCellAtlas/dcp-community/pull/58) PR?

diekhans:
This isn't replaced by the integration tests.  The integration tests are a way to test if other components are impacted by data this is entering the system.

This PR is about testing internally to metadata.  The kind of tests you would run before doing a commit or by CI.  

So I think it still has value.  We probably need to decide if we want to  keep this sitting as a PR for a while or move it forward and then wish for the resources to implement it.

docs/test_plan.md
malloryfreeberg: This would be brilliant to have :)

docs/test_plan.md
malloryfreeberg: One approach would be to generate JSON documents directly from the schema, which _should_ necessarily be valid :)

malloryfreeberg:
Merging this PR in now so we have this draft document available for review/updating at a later date.

Issue #572, opened 2018-10-04
Add protocol_type field to imaging_protocol.json


hewgreen:
**Give the bug a title**

protocol_type is a required field but the field is not in the schema. Worryingly this passed validation.

We either need to remove this from the required fields list or add the field.

**Schema versions**

imaging_protocol.json

**Expected behavior**

Is protocol_type a duplication of assay_type? We record assay type for each target.

        "assay_type": {
            "description": "Type of assay used to detect target.",
            "type": "object",
            "$ref": "module/ontology/process_type_ontology.json",
            "example": "MERFISH, smFISH, immunofluorescence, fluorescent cell stain",
            "user_friendly": "Assay type"
        },

As we have this do we need protocol_type at the level of imaging_protocol.json?

zperova:
There is no protocol_type in the original description of the imaging_protocol.json. So this must have migrated from copy-pasting during creating of the file?
https://docs.google.com/spreadsheets/d/1pQfWaywIK1e6YJFVdX5gFP7nwszD3Z6ZJppEVjbWA24/edit#gid=1087460193

zperova:
@hewgreen @dosumis if no one disagrees I will remove it.

malloryfreeberg:
Please remove it!

hewgreen:
I agree remove it. This is a bad way to describe what I presume it was anyway and capturing this at a more granular level is likely to be better.

PR #577, opened 2018-10-09
Remove protocol_type field  from imaging_protocol.json Fixes issue #572

zperova:
Remove protocol_type from imaging_protocol.json Fixes issue #572

Run the `src/schemas_are_valid_json.py` and `src/json_examples_validate_against_schema.py` scripts.

Changed update_log.


malloryfreeberg:
@zperova I think this is actually a bug fix. Given that the `protocol_type` field never existed in the schema, JSON documents that are valid against the old version of this schema wouldn't have the `protocol_type` field (would have been invalid) and thus they would still validate against this new version of the schema. Major version changes are reserved for backwards compatible breaking changes, which this change is not. Please update the update_log.csv file to reflect a "patch" change (aka a bug fix) and then I'll approve/merge. Thanks!

zperova:
Done! @malloryfreeberg

PR #588, opened 2018-10-12
Changed Project label to Project shortname

malloryfreeberg:
- Changed "Project label" to "Project shortname" in spreadsheet because the DCP integration test matches on the friendly name:

https://github.com/HumanCellAtlas/dcp/blob/master/tests/dataset_fixture.py#L42

`if project_tab['A2'].value != "Project shortname":`

It should instead match on the programmatic name:

`if project_tab['A4'].value != "project.project_core.project_short_name":`

A ticket is being made in dcp repo to fix this (thanks @samanehsan!). In the meantime, the spreadsheet is being changed so testing can continue.


samanehsan:
Thanks @malloryfreeberg! :octocat:

PR #622, opened 2018-11-19
Cell suspension schema patch updates.

malloryfreeberg:
For cell_suspension.json
- add user-friendly name to `biomaterial_core`
- remove example from `selected_cell_type`(example will live in the ontology module)
- remove example from `genus_species` (example will live in the ontology module)
- move "May be 1 for well-based assays." from description to guidelines for `total_estimated_cells` field. Also change to "Enter 1 for well-based assays."
- change "Total estimated cell count" to "Estimated cell count" for `total_estimated_cells` field


malloryfreeberg:
@daniwelter can you review and then do the merge if all looks ok to you? ty!

PR #631, opened 2018-11-23
JSON schema unit tests (Ready to merge)

simonjupp:
This PR contains the JSON schema unit tests rewritten to use the Elixir/HCA JSON schema validator. 

Please try and follow the instructions to run the tests. You will notice some of the tests are failing, these are genuine fails due to errors in either the schemas or the test data files.


daniwelter:
At least some of the failures are due to a mistake in links.json. I'm just doing a patch fix in that now.

daniwelter:
I merged the latest develop branch into this one as it contains a bug fix for the links.json schema. This resolves one error.

One potentially worrying failure of the schema validation is a case we came across recently: the validator doesn't pick it up if a field is listed as "required" but not actually present in the schema. I guess this is something we would need to implement as part of the linter.

daniwelter:
After some more debugging, I can confirm that all tests fail due to a validator issue rather than a problem with the example json. For some reason, after the first schema test, the validator doesn't appear to load any further schemas and tests all remaining files against the first schema . I replicated the problem with various schemas by reordering the test list.

simonjupp:
All tests are now passing so this could be a good time to merge this. These tests are more comprehensive than the existing python tests bu there is still more work to do on improving the tests and documentation

simonjupp:
All tests are now passing so this could be a good time to merge this. These tests are more comprehensive than the existing python tests bu there is still more work to do on improving the tests and documentation

PR #633, opened 2018-11-26
Changed technical_replicate_group in sequence_file.json rename to library_prep_id Fixes #262

### LINKED ISSUE #262: Rename technical_replicate_group_id to library_preparation_id
Not sure what this would go under but the issue is confusing labeling for library_preparation_process which leads to misunderstandings in library grouping. This should be an easy fix as its more of a current way we do it rather than a structure or content change to my understanding. We would just need to have a library_preparation_process for each library created. @malloryfreeberg would that work?

hewgreen:
Ran the src/schemas_are_valid_json.py and src/json_examples_validate_against_schema.py scripts locally. Updated update_log.csv

### Release notes

Changed technical_replicate_group in sequence_file.json rename to library_prep_id Fixes #262

### DIFF
--- a/json_schema/type/file/sequence_file.json
+++ b/json_schema/type/file/sequence_file.json
@@ -75,7 +75,7 @@
             "user_friendly": "INSDC run",
             "example": "SRR0000000"
         },
-        "technical_replicate_group" : {
+        "library_prep_id" : {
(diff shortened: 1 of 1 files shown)


hewgreen:
The example explains what a technical replicate is defined as and so was left in.

malloryfreeberg:
@zperova can you please re-review this PR and indicate if you are OK with the changes?

zperova:
all done

PR #635, opened 2018-11-26
Added two option fields for ENA submission Fixes #594

### LINKED ISSUE #594: Add nominal_length and nominal_sdev to library_preparation_protocol.json

I would like to request an update the the `library_preparation_protocol.json` schema.

I would like to add two new fields - `nominal_length` and `nominal_sdev` - to this schema to align with ENA standards and enable USI submissions to the archive. Open to alternative names for the fields, but it would be great if they were at least similar to the ENA fields to make it obvious they are representing the same value.

Field 1:

* Field name: nominal_length
* Field description: Average (insert) size of the fragments being sequenced.
* Required: no
* Example: 250
* CV or enum: no

Field 2:

* Field name: nominal_sdev
* Field description: Standard deviation of the (insert) size of the fragments being sequenced.
* Required: no
* Example: 30
* CV or enum: no

In the sequencingExperiment ENA metadata submitted via the USI, if the layout is paired, then the attributes NOMINAL_LENGTH and NOMINAL_SDEV are required. The HCA metadata standard should try to capture this metadata so it can be translated to ENA if needed (if `"paired_end": true`).

From Dave: 
> ENA require these fields for PAIRED layouts. Positive integers values should be fine, but present them as strings (i.e. "150" instead of 150)

Metadata appears in ENA Experiment XML like this:
```
               <LIBRARY_LAYOUT>
                   <PAIRED NOMINAL_LENGTH="250" NOMINAL_SDEV="30"/>
               </LIBRARY_LAYOUT>
```

Field definitions in ENA: 

>The attribute PAIRED NOMINAL_LENGTH is the average insert size. It is not the length of the reads. It is the average size of the fragments that are being sequenced.
>The attribute NOMINAL_SDEV is the standard deviation of the fragment lengths. This attribute is not mandatory so you can omit it if you do not have this detail.

hewgreen:
Ran schemas_are_valid_json.py and json_examples_validate_against_schema.py

Minor schema edit. Two new fields added to type/protocol/sequencing/library_preparation_protocol

Added new optional field nominal_length and nominal_sdev
```
"nominal_length": {
    "description": "Average (insert) size of the fragments being sequenced.",
    "type": "integer",
    "user_friendly": "Nominal length",
    "example": "250"
},
"nominal_sdev": {
    "description": "Standard deviation of the (insert) size of the fragments being sequenced.",
    "type": "integer",
    "user_friendly": "Nominal standard deviation",
    "example": "30"
```

### DIFF
--- a/json_schema/type/protocol/sequencing/library_preparation_protocol.json
+++ b/json_schema/type/protocol/sequencing/library_preparation_protocol.json
@@ -156,6 +156,18 @@
             "$ref": "module/ontology/library_amplification_ontology.json",
             "user_friendly": "cDNA library amplification method",
             "example": "PCR"
+        },
+        "nominal_length": {
+            "description": "Average (insert) size of the fragments being sequenced.",
+            "type": "integer",
+            "user_friendly": "Nominal length",
+            "example": "250"
+        },
+        "nominal_sdev": {
+            "description": "Standard deviation of the (insert) size of the fragments being sequenced.",
+            "type": "integer",
+            "user_friendly": "Nominal standard deviation",
+            "example": "30"
         }
     }
 }
--- a/json_schema/update_log.csv
+++ b/json_schema/update_log.csv
@@ -1,1 +1,3 @@
 Schema,Change type,Change message,Version,Date
+type/protocol/sequencing/library_preparation_protocol,minor,"Added new optional field nominal_length",,
+type/protocol/sequencing/library_preparation_protocol,minor,"Added new optional field nominal_sdev",,


hewgreen:
@malloryfreeberg can you provide these extra examples please?

Why do we need two examples for everything?

malloryfreeberg:
> Why do we need two examples for everything?

This is a point that needs to be clarified in the Style Guide. I think (ideally) Gabby wanted two examples for everything, but this PR/issue (and I can think of others) is an example of where adding a second example doesn't add much value to the schema. I propose the following:

- [ ] Keep only 1 example per field for this PR
- [ ] Update the Style Guide to include a "Special case" where adding a second example doesn't add value

For the second point, we'll need to figure out a way to word this in the Style Guide that isn't ambiguous.

zperova:
done

PR #636, opened 2018-11-26
Move normothermic_regional_perfusion to death module. Fixes #590.

### LINKED ISSUE #590: Move normothermic_regional_perfusion to death.json schema

I would like to request an update the the `donor_organism.json` and `death.json` schemas.

The field `normothermic_regional_perfusion` should be moved from the `donor_organism.json` schema to the `death.json` schema as it is only relevant if the donor is deceased. Usually it is done prior to organ donation.

No new fields requested.

This field was added to the `donor_organism.json` schema based on requirements for the Tissue Sensitivity dataset, but as it only applies in cases where the donor is deceased, it should be moved to the `death.json` module.

hewgreen:
Fixes #590 

Ran schemas_are_valid_json.py and json_examples_validate_against_schema.py

### Release notes 
major change to type/biomaterial/donor_organism.json: "Removed normothermic_regional_perfusion"
minor change to module/biomaterial/death.json: "Added normothermic_regional_perfusion"


malloryfreeberg:
@hewgreen @zperova @cjvillar can we please hold off on merging this one right now? There is a production dataset that uses this field and I don't want us to be sure that changing this field won't break anything in the DS or Browser in regards to using the affected dataset.

PR #645, opened 2018-11-27
Rename channel_name in imaging_target. Fixes #639

### LINKED ISSUE #639: Rename channel_name to channel_id in imaging_target.json

 I would like to request an update the `imaging_target.json` schema.

I would like to rename `channel_name` to `channel_id`.

The purpose of this field is to provide linking between targets and channels. The name `channel_id` fits better than `channel_name`.

zperova:
Rename channel_name to channel_id in imaging_target. Fixes #639

Ran `json_examples_validate_against_schema` and `schemas_are_valid_json` 

Major change - Changed name of an optional field.


hewgreen:
Some minor changes were broken down into quite a few PRs which was quite confusing. We shouldn;t break them down like this again. The conflicts problem with Zina's git hub made it more confusing. Here I ran the script, the update log went blank but git didn't detect changes! So I've no idea what went on. The metadata is good, we need to add an entry to the log, I've no idea if the versions were bumped.

malloryfreeberg:
> update log went blank but git didn't detect changes

I think this means that there was a mistake in the update_log.csv line. I think this happened to me once and so I re-wrote the update_log.csv, checked it, ran the release_prepare script again, and all was well.

In the future, we should do two things:
1. Update the release_prepare script to not delete the line from the file if the update doesn't happen
1. Not merge a PR if there is a schema change but no changelog.md, versions.json, and update_log.csv change.

We need to look at this in the morning and do a hotfix to the changelog and version.json file

daniwelter:
@malloryfreeberg @zperova Do you have any examples of update logs that failed? It would help debug this.

malloryfreeberg:
> Do you have any examples of update logs that failed?

If you enter a path to a schema that isn't a valid path, like `type/protocol/library_preparation_protocol`. I think that was the one that got me last time.

Also, maybe if you misspell major, minor, or patch? I think I had `pathc` once, and the script completed w/o errors but didn't actually update anything.

PR #675, opened 2018-12-07
Made development_stage required in schema donor_organism. Fixes #607

### LINKED ISSUE #607: Make development_stage required

donor_organism

Make the field development_stage a required field of the donor_organism schema

Donor age is currently used for faceting in the data portal. In the majority of cases, the development stage of the donor (eg adult, juvenile, pre-natal) is more relevant than the specific age, except for very specific cases such as early embryonic development stages. While age should still be capture when available, each donor should be assigned a development stage as a matter of course. If there is no indication that a donor is not an adult, adult stage should be accepted as a default.

This ticket came out of the semantic searching and browser sessions at the November DCP meeting.

daniwelter:
I have made development_stage a required property of donor_organism. This a major schema change that breaks backwards compatibility, so may require a hold on the PR merge.


malloryfreeberg:
@daniwelter what are all these changes in the tests/package-lock.json file? Look like version differences?

daniwelter:
I'm not sure, this seems to happen every time I run the new validator. If it still runs ok on Travis, I'd leave it be.

malloryfreeberg:
We should just merge this in. If we break backwards compatibility of stuff already in the DS, so be it. Will be a good use case if this causes issues down the line.

PR #686, opened 2018-12-14
Sequencing protocol patch updates.

malloryfreeberg:
### Release notes
- add user-friendly term to protocol_core
- remove "Should be a child term..." from instrument_manufacturer_model
- remove example from instrument_manufacturer_model (example will live in the ontology module)
- remove example from sequencing_approach (example will live in the ontology module)
- add second example for local_machine_name
- change description of paired_end to "Whether the sequenced molecule was sequenced from both ends."
- change approach to method in sequencing_approach desc and user-friendly

### Reviews
- Needs 2 Reviewers
- Second Reviewer to merge changes


zperova:
I think it is useful to have this somewhere for each ontology term: "Should be a child term of https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0000548" 
Otherwise when the wrong term is used it is only known upon upload of the spreadsheet.
I believe a good place for it will be the doc with instructions that @daniwelter was writing up?

malloryfreeberg:
> I think it is useful to have this somewhere for each ontology term

Agreed. This bit will live in the ontology schemas themselves, eventually. We need to figure out how best to format this, though, so it doesn't look overwhelming in the spreadsheet, UI, etc.

PR #751, opened 2019-01-18
Remove commas from storage_method enum list. Fixes #743.

### LINKED ISSUE #743: storage_method enum is unclear
**specimen_from_organism.preservation_storage.storage_method**

Enum for this field as per error message allows:

ambient temperature,cut slide,fresh,frozen, -70C freezer,frozen, -150C freezer,frozen, liquid nitrogen,frozen, vapor phase,paraffin block,RNAlater, frozen,TRIzol, frozen

This looks like repeated options due to comma fields in enum. Plus confusingly 'frozen' is not a valid option which is contrary to the example `For example: frozen, liquid nitrogen'.

This behaviour is not an ingest bug. See schema below.
```
"storage_method": {
            "description": "The method by which a biomaterial was stored after preservation or before another protocol was used.",
            "type": "string",
            "enum": [
                "ambient temperature",
                "cut slide",
                "fresh",
                "frozen, -70C freezer",
                "frozen, -150C freezer",
                "frozen, liquid nitrogen",
                "frozen, vapor phase",
                "paraffin block",
                "RNAlater, frozen",
                "TRIzol, frozen"
            ],
            "example": "frozen, liquid nitrogen; fresh",
            "user_friendly": "Storage method"
        }
```

malloryfreeberg:
### Release notes

For preservation_storage:
    - remove comma from enum list values
    - remove redundant word "freezer"
    - specify temperatures for RNAlater
    - remove Trizol as value (relevant for storing RNA, not specimen)

### Reviews
- Needs 2 Reviewers because this is a major version change
- Second Reviewer to merge changes


pnejad:
@malloryfreeberg  In my experience, we have sorted cells after dissociation and stored them in trizol. By not relevant for specimen preservation, do you mean the specimen immediately after collection?

malloryfreeberg:
> By not relevant for specimen preservation, do you mean the specimen immediately after collection

Yes, this set of fields is currently only attached to the specimen, which is the biomaterial that is taken directly from the donor. Typically, this is a bit of tissue. 

If we want to extend this set of fields to also be used to describe how cells are preserved or stored, we can decide whether it makes sense to just keep adding to the enum or whether we need a new field.

PR #762, opened 2019-01-22
[on hold] Initial draft of the schema linking file

daniwelter:
This is not a schema change as such but rather a proposal for a new file called `schema_linking.json` that captures two things:
1. Each biomaterial chain we have encountered so far in a dataset (`biomaterial_linking`)
2. All possible locations for protocol links (`protocol_linking`

The reason I'm proposing this file is to allow us to automatically build comprehensive testing spreadsheets for all possible graph shapes (for biomaterial chains) with maximally complete protocol sets. At the moment, the only way to automatically insert linking fields into the spreadsheet is to include them in a yaml file. If a new schema is added, this would mean updating all the yaml files for all the possible spreadsheet scenarios, as well as having to generate a new yaml file every time we encounter a new biomaterial chain.

The proposed linking file will live in the metadata repo and can be updated by any metadata editor. If a new type schema is added or a new graph shape encountered, this should be captured in the file. This will automatically generate a new spreadsheet for the integration test suite.

Comments on the plan in general and the overall structure of the file please.


hewgreen:
Biomaterial chains may be all you need for your tests and ti build spreadsheets but for later stats and analysis a few things would be useful to add if you could:

Frequency of use in the system
Links to datasets or names
Protocol variants
Number of entities at each step in the chain

'Not right now' seems like an appropriate response to this but if we were to capture these things would this be a good place for that?

json_schema/schema_linking.json
```diff
@@ -0,0 +1,119 @@
...
+    },
+    {
+      "specimen_from_organism": "donor_organism",
+      "cell_line": "specimen_from_organism",
+      "cell_suspension": "cell_line",
+      "sequence_file": "cell_suspension"
+    },
+    {
+      "specimen_from_organism": "donor_organism",
+      "cell_line": ["specimen_from_organism", "cell_line"],
```
malloryfreeberg: What happens here? An example is made for each of the values in the array? Or would this block produce the example DO -> SFO -> CL -> CL -> CS -> SF? How would the code know to make the second CL -> CL link?

daniwelter: The code isn't written yet, so it'll do whatever I tell it to do. In this instance, if presented with an array, it would create a linking column in the cell_line tab to specimen.id and one to cell_line.id.

json_schema/schema_linking.json
```diff
@@ -0,0 +1,119 @@
...
+    ],
+    "differentiation_protocol": [
+      {
+        "source": "cell_line",
+        "output": "cell_suspension"
+      },
+      {
+        "source": "cell_line",
+        "output": "organoid"
+      }
```
malloryfreeberg: I think technically you can differentiate a cell line into another cell line? @pnejad ?

json_schema/schema_linking.json
```diff
@@ -0,0 +1,119 @@
...
+    ],
+    "dissociation_protocol": [
+       {
+        "source": "specimen_from_organism",
+        "output": "cell_line"
+      },
```
malloryfreeberg: Can one actually dissociate a cell line from a specimen?

json_schema/schema_linking.json
```diff
@@ -0,0 +1,119 @@
...
+      {
+        "source": "specimen_from_organism",
+        "output": "cell_suspension"
+      }
```
malloryfreeberg: I think one can also dissociate a cell suspension from an organoid and from a cell line

json_schema/schema_linking.json
```diff
@@ -0,0 +1,119 @@
...
+        "output": "cell_line"
+      },
+      {
+        "source": "specimen_from_organism",
+        "output": "cell_suspension"
+      },
+      {
+        "source": "cell_line",
+        "output": "cell_suspension"
+      }
```
malloryfreeberg: And cell suspension from an organoid.

malloryfreeberg:
Do we need to consider analysis outputs at all?

malloryfreeberg:
Are all 13 of Matt's graph shapes captured in this file?

json_schema/schema_linking.json
daniwelter: For the first pass, I captured all the scenarios listed here https://github.com/HumanCellAtlas/hca-data-wrangling/blob/master/docs/SOP_docs/HCA_Metadata_Spreadsheet_Guide_Wranglers.md#adding-in-protocols

daniwelter:
> Do we need to consider analysis outputs at all?
As far as I'm aware, there are currently no plans for an analysis tab in the spreadsheet

> Are all 13 of Matt's graph shapes captured in this file?
Simon, Matt and I made the decision on Monday not to capture the full graph shapes but instead separate out the biomaterial derivations and the associated protocols. The file captures all the biomaterial derivations we've seen to so far and the plan is to make the protocols for these maximally complete, ie every possible protocol for a given chain based on the protocol_linking rules are included.

json_schema/schema_linking.json
zperova: yes, neuro_diff is an example

json_schema/schema_linking.json
zperova: you can dissociate cells from specimen to make a cell line, so - yes

malloryfreeberg:
> As far as I'm aware, there are currently no plans for an analysis tab in the spreadsheet

Sure, but I'm wondering if we should also be generating fake analysis JSON docs for the purposes of having a complete set of example JSONs.

json_schema/schema_linking.json
daniwelter: Added.

json_schema/schema_linking.json
daniwelter: Added

json_schema/schema_linking.json
daniwelter: Added

zperova:
I assume this will have to change to include the library_prep once we decide on it?

daniwelter:
Closing this pull request as it is unclear whether this feature is still relevant.

PR #769, opened 2019-01-24
Change familial_relationship to be plural (array). Fixes #768.

### LINKED ISSUE #768: Make familial_relationship plural to adhere to Style Guide.

I would like to request an update to the `donor_organism.json` schema.

- [x] change familial_relationship to familial_relationships since the field takes an array

No new fields.

Would like to adhere to the Style Guide.

malloryfreeberg:
### Release notes

For donor_organism.json schema:
- change familial_relationship to familial_relationships since the field takes an array

### Reviews requested


json_schema/type/biomaterial/donor_organism.json
```diff
@@ -122,10 +122,11 @@
             "$ref": "module/biomaterial/death.json",
             "user_friendly": "Death conditions"
         },
-        "familial_relationship": {
+        "familial_relationships": {
             "description": "Information about other organisms related to this organism.",
             "type": "array",
             "items": {
+                "type": "object",
```
daniwelter: Is this really necessary? Json refs are technically supposed to be the only property in an object as per JSON schema spec as Simon and I discovered yesterday. As a result, ref importing overwrites any sibling fields to $ref (from https://json-schema.org/latest/json-schema-core.html#rfc.section.8.3 - "All other properties in a "$ref" object MUST be ignored."). Technically including other fields is perfectly valid JSON but they will never be considered by a validator.

malloryfreeberg: Interesting. I was just explicitly stating type: object to keep everything consistent with the style guide (where type is a required attribute).

malloryfreeberg: On second thought, let's keep this clean and not have any other attributes for properties that use $ref. I'll remove it here, make this explicit in the Style Guide, and we can update other fields later.

malloryfreeberg:
@willrockout could you please review and merge this PR? Thanks!

PR #773, opened 2019-01-30
Added a printed warning for update types

daniwelter:
As per recent discussion about the use of change vs changed etc, I've added a warning to the release script if the type isn't part of a defined list. The script will still pass but you should see a warning

### DIFF
--- a/src/release_prepare.py
+++ b/src/release_prepare.py
@@ -81,6 +81,9 @@ class ReleasePrepare:
                 version = log_content[val][self.version_column]
                 type = change_message.split(' ')[0]
 
+                if type not in ['Added', 'Changed', 'Removed', 'Fixed', 'Deprecated', 'Security']:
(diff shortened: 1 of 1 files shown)


malloryfreeberg:
Can the script, instead of just throwing a warning, actually exit without doing any changes? This way the author has the chance to update the update_log.csv with the correct wording and then re-run the release_prepare script again.

zperova:
This will be very helpful!

daniwelter:
Of course it can - are you absolutely sure that's what you want?

malloryfreeberg:
> Of course it can - are you absolutely sure that's what you want?

I think so. If the condition is met, throw the warning to the screen and also exit the script w/o doing the preparation. That allows us to address the warning and then re-run the script instead of having to manually update the changelog.

daniwelter:
Ok, done. Please review and merge at your leisure :)

PR #833, opened 2019-02-20
New property migrations json file

daniwelter:
### Release notes

No schema changes.

New `property_migrations.json` file capturing the latest round of major schema changes. 
New `sample_property_migrations.json` providing examples of types of property migrations.
New `json_doc_migrations.py` script providing proof of concept migrations using the property migrations json file and several migration test json docs.

### Reviews requested

Tagging @hannes-ucsc and @simonjupp as an FYI. Merging should not affect the repo content so can be merged any time.


json_schema/property_migrations.json
```diff
@@ -0,0 +1,60 @@
...
+    {
+      "source_schema": "cell_suspension",
+      "property": "total_estimated_cells",
+      "target_schema": "cell_suspension",
+      "replaced_by": "estimated_cell_count",
+      "effective_from" : "9.0.0",
+      "reason": "Field clarification"
+    },
+    {
+      "source_schema": "library_preparation_protocol",
```
malloryfreeberg: @mckinsel in response to your question about guidance for working with schema migrations, here is an example of an approach being developed that will provide direct guidance for field-by-field migrations.

json_schema/sample_property_migrations.json
```diff
@@ -0,0 +1,48 @@
...
+    {
+      "source_schema": "specimen_from_organism",
+      "property": "biomaterial_core.biomaterial_id",
+      "target_schema": "specimen_from_organism",
+      "replaced_by": "specimen_id",
+      "effective_from" : "8.0.0",
+      "reason": "Schema demodularisation"
+    },
+    {
+      "source_schema": "specimen_from_organism",
```
malloryfreeberg: What schema migration is this block representing? I see the first two are field name changes/moves.  I'm not sure what this third one is saying.

malloryfreeberg: Why is this change recorded as in the `specimen_from_organism` schema when the change happens in the `disease_ontology` schema?

daniwelter: Because the migrations are encoded against submittable types, which is where they'll need to be dereferenced against. The downside is that a migration in a widely-used module will have to be encoded for every submittable type it touches. The upside is that you can migrate json docs directly against a migration without having to fully introspect the source schemas.

malloryfreeberg:
I'm still a little nervous that it's not immediately clear what the type of change is for each block. Would it not be better to have an attribute like `"type"` where it's directly stated whether the change is, e.g., `"type": "deprecated property"` or `"type": "moved property"` or `"type": "new required property"`?

daniwelter:
@malloryfreeberg no objections - type now added to sample and actual property migrations. It does mean an additional field to fill in for metadata schema editors in the short-term, unless we get rid of "reason".

malloryfreeberg:
The "reason" seems like a "nice-to-have". One could theoretically look up the reason in the changelog or the PR or the GH issue. Alternatively, the "reason" could automatically be filled in using the line entered in the update_log.csv. We'd have to move towards a more standard way of writing that line, though, to include the reason.

daniwelter:
Agreed, and I'm happy adapt the reason and type fields as we go along. I can even take care of filling in the migrations for everyone for now, until we have a proper SOP and/or tooling in place. I just really need the properties file the schema server for now.

PR #840, opened 2019-02-22
Remove outputs field from analysis_process schema. Fixes #664.

### LINKED ISSUE #664: analysis_process mis-uses submittable type as module
The property `outputs` references a submittable type as if it were a module. This causes issues with the new JSON Schema validator as the injected fields are treated as a separate submittable type, causing failures.

malloryfreeberg:
### Release notes

For analysis_process schema:
- Removed required inputs and outputs fields. Fixes #664. 

### Reviews requested

- Need 2 Reviewers to approve because this is a major update.
- Need @rexwangcc to approve because pipelines team will need to update some code.
- Last Metadata Team member to review/approve, merge changes by Friday March 1.


tests/schema_test_files/process/test_pass_analysis_process.json
```diff
@@ -4,353 +4,6 @@
   "input_bundles": [
     "f3942c14-2024-4756-9360-92d89d515c9b"
   ],
-  "inputs": [
```
malloryfreeberg: @rexwangcc even though ingest doesn't use this information to build links between the inputs and outputs of the analysis, looking at the values in the `inputs` field makes me think that it might actually be useful for people. What do you think? The "fastq1" and "fastq2" information is actually fine here, schema-wise. Unlike for the `outputs` field, the `inputs` field is not actually importing the `analysis_file.json` _Type_ schema, so it would be ok to keep the `inputs` field.

diekhans:
This very much relates to something I am about to write up involving the metadata graph and versioning.  I will put that in another ticket.

The summary is that I think this change should go in and a more general way to represent provenance of the metadata neeeds to be developed.

tests/schema_test_files/process/test_pass_analysis_process.json
rexwangcc: That makes sense to me!! That said, you might want to bring the `inputs` back in this test file?

tests/schema_test_files/process/test_pass_new_analysis_process.json
```diff
@@ -26,27 +26,5 @@
...
-      "file_core": {
-        "file_format": "bam",
-        "file_name": "03beed16-8eae-4fa4-997b-cea1dddee4a7_qc.bam"
-      },
-      "schema_type": "file",
-      "describedBy": "http://schema.dev.data.humancellatlas.org/type/file/5.3.4/analysis_file"
-    }
-  ],
   "analysis_run_type": "copy-forward"
 }
```
rexwangcc: nit: new line at the end

tests/schema_test_files/process/test_pass_analysis_process.json
malloryfreeberg: Will do!

PR #879, opened 2019-03-08
Changed filter_range example in channel.json. Fixes Issue878

zperova:
### Release notes

For channel.json schema:
- Fixed example in filter_range and added guidelines.
 

### Reviews requested

- Need 1 Reviewer to approve because is a patch update


hewgreen:
Do we have evidence that CWL/BW is the best format? Do other archives use this? Also I wondered if we need to specify FWHM or HBW rather than just BW?

zperova:
The CWL/BW is the standard format you get from the manufacturer, so is easy for the contributor to supply. @hewgreen

PR #881, opened 2019-03-11
Added second example to test_results. Fixes #846.

### LINKED ISSUE #846: Add second example to test_results

I would like to request an update to the `medical_history.json` schema.

I would like to add a second example - "HIV, HBV, HCV: Negative" - to the `test_results` field.

No new fields.

Would like to adhere to Style Guide stating that - where appropriate - two examples should be provided in the schema.

malloryfreeberg:
### Release notes

For medical_history.json schema:
- Added second example to test_results. Fixes #846.
 
### Reviews requested

- Need 1 Reviewers to approve because this is a patch update
- Last to review/approve, merge changes by **Thursday March 14**.


malloryfreeberg:
@willrockout It looks like both schemas were incremented by 2 patch versions. This is unusual as the schemas should have only incremented by 1 version. Did you happen to run the release_prepare script twice? It would be good to understand how this happened to be sure there wasn't a mistake made anywhere.

willrockout:
@malloryfreeberg Looking back through my logs I see that I did run it twice but the first one errored out because I ran python instead of python3. It must have incremented the versions file and not removed the update_log file then the second run incremented it again.

malloryfreeberg:
Ok. I don't think this is technically an issue (there just won't be those two skipped released versions in the official standard). Moving forward it would be good to have a mechanism in place for preventing this from happening. 

Perhaps the easiest thing to do right now is to add a line in the release process doc about what to do if the release_prepare script fails for any reason. The answer is to use git to remove all the changes that were done and then re-run the release prepare script. This would have removed the first patch increment changes and reset back to the current versions. I'll make a ticket to add this documentation.

PR #886, opened 2019-03-14
Probe module to replace target module. Fixes Issue#813.

zperova:
### Release notes

New probe.json schema contains the following fields (* - required):
- probe_label*
- probe_sequence
- target_codebook_label
- target_label*
- subcellular_structure
- manufacturer
- catalog_number
- lot_number
- assay_type*
- fluorophore
- channel_label
- codeword
 
### Reviews requested

- Need 2 Reviewers to approve because this is a major change


json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "example": "AGGCTATAGCGGAGCTACG; aggctatagcggagctacg"
+        },
+        "target_codebook_label": {
+            "description": "A label used in the codebook for the target.",
+            "type": "string",
+            "example": "AKT1; CFOS",
+            "user_friendly": "Target label in codebook",
+            "guidelines": "Leave blank if experiment did not use codebook. "
+        },
+         "target_label": {
```
hewgreen: This should not be called label as it is not the primary key. It should also be an ontology field.

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "description": "An identifier for the target molecule.",
+            "type": "string",
+            "example": "CHEBI:85345; ENSG00000170345",
+            "user_friendly": "Target label",
+            "guidelines": "For small molecules this should be from the ChEBI ontology. For gene products this should be a standard gene or gene product identifier from e.g. Ensembl or UniProt."
+        },
+        "subcellular_structure": {
```
hewgreen: Can we remove this field if this can be covered by ontologies under the field currently called 'target_label'?

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "description": "Target subcellular structure. This should be a term from the GO cell component ontology.",
+            "type": "object",
+            "$ref": "module/ontology/cellular_component_ontology.json",
+            "user_friendly": "Target subcellular structure"
+        },
+        "manufacturer": {
```
hewgreen: I thought these fields would come from the purchased reagent module?

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+        "probe_label": {
+            "description": "The label of a probe used to detect target in this experiment.",
+            "type": "string",
+            "example": "ACTA1; cFos",
+            "user_friendly": "Probe label"
+        },
+         "probe_sequence": {
```
hewgreen: I think this needs to be more specific so contributors only provide the sequence that binds to the target and not include sequences of adaptors etc.

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "user_friendly": "Batch/lot number"
+        },
+        "assay_type": {
+            "description": "Type of assay used to detect target.",
+            "type": "object",
+            "$ref": "module/ontology/process_type_ontology.json",
+            "example": "MERFISH; in situ sequencing",
+            "user_friendly": "Assay type"
+        },
+        "fluorophore": {
```
hewgreen: Just want to double check that this doesn't need to be an array. I know the probe and secondary probes can be quite different such as the case in RNAscope.

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "description": "Channel ID used to assay signal. Should be consistent with the ID in the channel tab.",
+            "type": "array",
+            "items": {
+                "type": "string"
+            },
+            "example": "1; A",
+            "user_friendly": "Channel",
+            "guidelines": "Should be the same as in Channel tab."
+         },
+        "codeword": {
```
hewgreen: Add another example. Do you think it's worth adding a MERFISH example?

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "user_friendly": "Assay type"
+        },
+        "fluorophore": {
+            "description": "Fluorophore used to detect target.",
+            "type": "string",
+            "example": "Cy5",
+            "user_friendly": "Fluorophore"
+        },
+        "channel_label": {
+            "description": "Channel ID used to assay signal. Should be consistent with the ID in the channel tab.",
```
hewgreen: Change Channel ID to channel label in the description.

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Information about probes used to detect targets.",
+    "additionalProperties": false,
+    "required": [
+        "probe_label",
+        "probe_sequence",
```
hewgreen: Probe sequence cannot be required for antibody based methods on protein targets etc.

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+            "description": "Type of assay used to detect target.",
+            "type": "object",
+            "$ref": "module/ontology/process_type_ontology.json",
+            "example": "MERFISH; in situ sequencing",
+            "user_friendly": "Assay type"
+        },
+        "fluorophore": {
+            "description": "Fluorophore used to detect target.",
+            "type": "string",
+            "example": "Cy5",
```
hewgreen: needs another example

hewgreen:
@simonjupp the two nested imaging arrays probe and channel are currently modules but ideally just need nesting in imaging protocol. Should we do this now or are we waiting for something else to happen?

dosumis:
I still don't know what the codeword is doing in here - given that this is redundant with the SpaceTx/starfish codebook.

hewgreen:
Codeword is an optional field but contributors doing in-situ sequencing have added it when filling out the metadata. We could delete this information or add the field and track usage. I'd be happy either way because as you say the information can be deduced from the codebook.

json_schema/module/protocol/probe.json
daniwelter: This can't be an ontology field if it needs to accept both Chebi IDs and Ensembl gene or Uniprot IDs.

json_schema/module/protocol/probe.json
daniwelter: @zperova tried to explain this use case to me but it doesn't make any sense as far as I'm concerned. Chebi isn't axiomatised with GO, so you can't infer cellular location from a Chebi ID. Uniprot and Ensembl are annotated with GO terms but as they are not ontologies themselves, we can't leverage inference here either.

json_schema/module/protocol/probe.json
daniwelter: +1 on purchased reagents module here, the next two fields also fall into this category.

zperova: purchased reagent is a module so cannot be inside a probe module. This makes the structure too nested.  @hewgreen @daniwelter

zperova: this is even more relevant if we are going down schema demodulirisaton route

json_schema/module/protocol/probe.json
zperova: they are not called adaptors. I avoided using the terms that can be confusing, and thought that we can update the description as a patch update once the schema is in

json_schema/module/protocol/probe.json
zperova: good catch, @hewgreen 
here is another use case for conditional validation

json_schema/module/protocol/probe.json
zperova: you will only have one probe to one fluorophore correspondence in RNAscope. the primary probe is not conjugated to a flurophore

json_schema/module/protocol/probe.json
zperova: i intend this to be only used in ISS-based experiments

json_schema/module/protocol/probe.json
daniwelter: By that argument, we also can't use ontology annotations in modules, which we do in 7 different instances. This existing target.json module imports purchased_reagents.

json_schema/module/protocol/probe.json
malloryfreeberg: > This should not be called label as it is not the primary key.

I don't think there is a rule that only primary key fields can contain `_label`. I am OK with this particular field having `_label` in it. @hewgreen can you suggest another suitable term if you don't like "label"?

json_schema/module/protocol/probe.json
malloryfreeberg: The description and guidelines attribute are perfect to use here to specify that only the sequence that directly recognizes/binds to the endogenous target should be provided here. I don't think we can convey this type of information in the field name alone. @zperova I would make this distinction clear here instead of waiting until the schema is released and then doing a patch update.

json_schema/module/protocol/probe.json
```diff
@@ -0,0 +1,109 @@
...
+        },
+         "target_label": {
+            "description": "An identifier for the target molecule.",
+            "type": "string",
+            "example": "CHEBI:85345; ENSG00000170345",
+            "user_friendly": "Target label",
+            "guidelines": "For small molecules this should be from the ChEBI ontology. For gene products this should be a standard gene or gene product identifier from e.g. Ensembl or UniProt."
+        },
+        "subcellular_structure": {
+            "description": "Target subcellular structure. This should be a term from the GO cell component ontology.",
```
malloryfreeberg: > This should be a term from the GO cell component ontology.

This phrase should be moved to the guidelines attribute or removed. We don't have directions for other ontologized fields at the moment (waiting for lookup tooling), but if you want to keep this in, move it to guidelines.

json_schema/module/protocol/probe.json
malloryfreeberg: @daniwelter we currently make exceptions for ontology modules. Otherwise, no other modules currently import other modules. If we are moving towards demodularization anyway, I think it's ok to flatten here and just include the fields from the `purchased_reagents` module that are relevant here. @zperova indicated that not all of the `purchased_reagents` module fields are relevant.

json_schema/module/protocol/probe.json
malloryfreeberg: @hewgreen can you suggest another example?

json_schema/module/protocol/probe.json
malloryfreeberg: Also remove "Should be consistent with the ID in the channel tab." as this bit is covered in the guidelines attribute.

json_schema/module/protocol/probe.json
hewgreen: You can use the term adaptor to mean this https://en.wikipedia.org/wiki/Adapter_(genetics)

json_schema/module/protocol/probe.json
hewgreen: This should not be a module anyway. We should just nest this directly in the main type. This was the most obvious case for reducing module usage.

json_schema/module/protocol/probe.json
hewgreen: The most important use case is for a user to provide 1 intended target of 1 probe. E.g. DAPI = DNA, ATGC... = a gene (or transcript), Antibody = a protein. Therefore, I see this as one field 'intended probe target' with values ranging in granularity from subcellular structures (for nonspecific binders) to specific elements. As two fields they are mutually exclusive and so should be one.

json_schema/module/protocol/probe.json
hewgreen: @malloryfreeberg name will depend on the solution to the technical problem @daniwelter highlights.

json_schema/module/protocol/probe.json
hewgreen: If it isn't conjugated could a coding strategy involve different fluorophores binding to the same probe in different rounds?

json_schema/module/protocol/probe.json
hewgreen: I'm not sure why you would want to specify that (although as discussed this is not an optimal format for other experiments), it may be used for non ISS. However, if you want to this should be clearly stated in guidelines or description.

json_schema/module/protocol/probe.json
hewgreen: Cy2, Cy3, Cy7, FITC

zperova: i left one example as teh second doesn't really add value (as they are essentially the same)

json_schema/module/protocol/probe.json
zperova: as @daniwelter mentions, this can not be an ontology field. @hewgreen what is your suggestion for an non-ontologized not primary key field?

json_schema/module/protocol/probe.json
zperova: @hewgreen do you propose to get rid of sub-cellular structure taken that we can not incorporate it at the moment in any other field?

json_schema/module/protocol/probe.json
zperova: @hewgreen @daniwelter @malloryfreeberg since probe.json is a module at the moment and we do not need all the fields in the reagents module (most of them are left empty by contributor anyway), I prefer to include the ones I indicated to make it easier for the contributor to fill in.

json_schema/module/protocol/probe.json
zperova: as in MERFISH? yes, this will be reflected in the codebook then.
This field is for the one-to-one correspondence in the non-multiplexed experiments.

json_schema/module/protocol/probe.json
zperova: description says "sequencing cycles", you think this is not clear enough @hewgreen 
I thought we have agreed on this in our discussion about the MERFISH code being sparse and this field being user-suggested

zperova:
I updated guidelines and examples, removed codeword and made probe_sequence not required.

json_schema/module/protocol/probe.json
zperova: @hewgreen I asked people who do experiments, they don't call them adapters. they call them different things depending on the method and type of probe.

json_schema/module/protocol/probe.json
zperova: @malloryfreeberg done

json_schema/module/protocol/probe.json
hewgreen: Not sure I follow this logic. Are you talking about duplicating fields?

json_schema/module/protocol/probe.json
hewgreen: @daniwelter Is there no way to ontologies a field so that it can describe various biological molecules?

json_schema/module/protocol/probe.json
hewgreen: The fluorophore is not captured in the codebook.

Sounds like this field is for non multiplex only (and should be described as such). Or made into an array. This could be removed and the channels plus codebook could be used to work out the correspondence.

json_schema/module/protocol/probe.json
hewgreen: Does this conform to the style guide? Couldn't we use this argument for most second examples?

json_schema/module/protocol/probe.json
hewgreen: 'sequencing cycles' is too cryptic for my liking. Especially for experimenters unfamiliar with ISS who will see this field.

json_schema/module/protocol/probe.json
hewgreen: The 'target_label' discussion needs to be resolved before we can discuss this field.

json_schema/module/protocol/probe.json
malloryfreeberg: @zperova I would just add "FITC" and call it done. This is a borderline case for "adding a second example doesn't add value", but as it's very simple/easy to just add another example, I would do it. 
`"example": "Cy5; FITC",`

json_schema/module/protocol/probe.json
zperova: @hewgreen why would I want to duplicate fields? I suggest to add only the fields that are needed instead of using the purchased_reagents module that has the needed fields and also the extra fields that won't be filled in. I think this is a simpler solution.

json_schema/module/protocol/probe.json
zperova: @hewgreen what is your suggestion?

json_schema/module/protocol/probe.json
zperova: @malloryfreeberg I added another example

json_schema/module/protocol/probe.json
daniwelter: We can absolutely use multiple different ontologies to a given ontology field, we do that already for a range of ontology modules, but several of the anticipated identifiers (Uniprot, Ensembl) aren't from ontologies. Sure, the ontology annotation itself can be optional but that's not really very useful. Without a good range of examples values to assess coverage across the different resources, it's hard to make a call on this.

json_schema/module/protocol/probe.json
zperova: changed

json_schema/module/protocol/probe.json
hewgreen: Gathering free text to gain the examples we need is quite risky as there will potentially be lots of targets and the mapping may be impossible. 

How about two broad fields?
1. a target id field to take identifiers.org compliant IDs (inc protein isoforms, ensemble etc)?
2. a target ontology field that takes terms for subcellular entities, genes, proteins (pr) and any other ontologies @daniwelter thinks may be useful to describe molecular targets.

json_schema/module/protocol/probe.json
hewgreen: I didn't know you could import one field from a module.

json_schema/module/protocol/probe.json
hewgreen: I wouldn't make it ISS specific at this early stage. This field is not required as the information is in the codebook. We were including it to see how submitters used it as some already gave us this information without the field.

json_schema/module/protocol/probe.json
zperova: I do not think it is. I think the purchased_reagent module has to be reconsidered, so using it just because it is used in other schemas at the moment is not strong enough of an argument when we can do something better for the imaging schema right now.

json_schema/module/protocol/probe.json
zperova: After a quick look into what the search for the targets in one codebook gives, many targets are in the Cell Cycle Ontology. However this might be just by chance. Would it be much worse to ask the contributor to indicate that the target it either nuclear or cytoplastic - and give a choice of two ontology terms rather then a multitude since we capture the target_label in a different field? @daniwelter what do you think?

daniwelter: Before we massively extend the list of HCA-compliant ontologies willy nilly to include ontologies such as PR and CCO that are completely out of our control, I'd really like to do a coverage assessment of expected values to identify the best ontologies for our use cases. Can anyone provide me with a list of values we have so far?
@zperova in your 2-value use case, an enum would probably be more appropriate than an ontology field.

zperova: I will be happy with enum for now. @hewgreen ?
@daniwelter do you think that the targets also have to be ontologized if we are using Uniprot, Ensemble identifiers? I do not believe that whatever examples we have now are representative samples for use cases, but I will send you the list

daniwelter: @zperova No, I specifically don't want ontologised fields if we primarily expect things like Uniprot and Ensembl identifiers. While both of these are stable, official identifiers for their relevant concepts, they are not ontology terms and we should avoid mixing and matching ontology-annotatable terms and terms we know can't be annotated in a given field.

hewgreen: Definitely not an enum. I prefer splitting this into ID fields and an ontology field. If Dani can't work out the ontology stuff at this moment in time we could make it free text for now and see how it is used.
In other places we have separate fields for different ID types. I'm not sure this is great practice and would sooner use identifier compliant IDs so we can have one field to cover protein and nucleic acid targets.

json_schema/module/protocol/probe.json
daniwelter: We seem to be operating under a misconception here that if you import a module, you have to use all the fields of that module. This was never the case. Modules group a set of related fields that may be used together, but not all fields in a module will necessarily apply to all instances where that module is used. In this instance, you haven't included retail_name, expiry_date and kit_titer. While I can see why kit_titer wouldn't be appropriate for this use case, I don't see why expiry_date is inappropriate here. 
If we're moving away entirely from modules, then by all means, do as you please, but if we're still using modules at the moment, I think it would be wrong to cherry pick where we import modules and duplicate fields instead.

zperova: @daniwelter the reason not to use the module here was that it becomes too nested as probes.json is already a module. As an alternative, I only included the fields that the contributor is likely to fill out. all of the fields are relevant but the truth us no one fills them out. I will import the module then in this instance until work on de-modularization and specific modules can be prioritised.

PR #958, opened 2019-04-18
Changed file_format to format. Fixes #375.

### LINKED ISSUE #375: Rename field file_format to format
Ping Hannes, Dave

### Original request

1. <file_format>
**Which schema needs to be changed?**
Change to file_core.json
**What field in that schema need to be changed?**
file_format field
**What should the new field name be?**
Change to format
Just to shorten things.

### Consequences

1. “format” is a reserved keyword in JSON schema. While keywords can be overloaded, this should be avoided if at all possible to prevent confusion over usage. “format” could also be used in other contexts, eg submission format, which would further overload the label.

1. From a usability point of view, this doesn’t comply with the principles for: “recognition rather than recall”, “visibility of system status”, “error prevention” [https://www.nngroup.com/articles/ten-usability-heuristics/](https://www.nngroup.com/articles/ten-usability-heuristics/) , as it might be ambiguous if not carrying the specific context (i.e. format could be for time, date, etc)

1. Will break analysis pipelines (writing analysis results to ingest; querying on input fastq files) and data portal filters.

malloryfreeberg:
### Release notes

For file_core.json schema:
- Changed file_format to format. Fixes #375.
 

### Reviews requested

- Last Reviewer to review/approve, merge changes by **Monday April 29**


json_schema/core/file/file_core.json
```diff
@@ -4,7 +4,7 @@
     "additionalProperties": false,
     "required": [
 	    "file_name",
```
ESapenaVentura: According to the style guide, shouldn't we consider changing also `file_name` to `name`? I don't know if `name` is a special field name and we can't use it, but I wanted to bring it up just in case.

malloryfreeberg: This is a good point. I'm worried that since we don't have a GitHub issue at the moment about this, we haven't had any formal discussion about renaming. I'm also of the personal opinion that the word "filename" is one word (I didn't agree with adding the underscore between "file" and "name"). We should save this change for the future.

PR #960, opened 2019-04-18
Changed protocol_type/process_type properties to type. Fixes #931.

### LINKED ISSUE #931: Change protocol and process `type` fields.

I would like to request an update to multiple schemas (see below).

- [x] analysis_protocol.json
    - [x] change `protocol_type` to `type` (actually to something else that is more descriptive than 
- [x] protocol.json
    - [x] change `protocol_type` to `type` to remove redundancy ("protocol")
- [x] process.json
    - [x] change `process_type ` to `type` to remove redundancy ("process")
- [x] analysis_process.json
    - [x] change `process_type ` to `type` to remove redundancy ("process")

Above.

Adhere to Style Guide.

    Ping Marcus, Brian, Saman when PR is made

ESapenaVentura:
### Release notes
For `analysis_protocol.json` schema:
- Changed `protocol_type` to `type`

For `protocol.json` schema:
- Changed `protocol_type` to `type` to remove redundancy ("protocol")

For `process.json` schema:
- Changed `process_type ` to `type` to remove redundancy ("process")

For `analysis_process.json` schema:
-  Changed `process_type ` to `type` to remove redundancy ("process")
 

### Reviews requested

- Last Reviewer to review/approve, merge changes by 29/04/2019


malloryfreeberg:
Pinging @mckinsel @brianraymor @samanehsan that these changes are being prepared to go into dev.

@samanehsan I think we'll have to coordinate the release of `analysis_process.type` and `analysis_protocol.type` to integration at the same time.

samanehsan:
Thanks @ESapenaVentura and @malloryfreeberg! I'll make a ticket to update the fields in secondary analysis.

We are currently using the `analysis_protocol` version 8.0.7 and `analysis_process` version 9.0.0: https://github.com/HumanCellAtlas/pipeline-tools/blob/master/adapter_pipelines/ss2_single_sample/adapter_example_static.json#L17-L18

Could we update to the new schema versions after the changes are promoted? Or will ingest no longer support the versions that we are currently using?

malloryfreeberg:
@samanehsan I _think_ ingest validates JSON documents against the latest schema for a given environment. In the integration env, for example, the latest versions are:
`analysis_protocol`: 8.0.7 
`analysis_process`:  9.0.0

which match what you're using. Once we deploy schema changes to integration, then ingest will use those latest versions (which will increment by 1 major version each).

It's unclear to me whether ingest will allow earlier versions to be asserted and used. @aaclan-ebi or @simonjupp or @justincc  do you know?

justincc:
I'm pretty sure that ingest will currently only allow latest versions to be used. Otherwise we wouldn't have the issue with updates only validating against the latest schema versions.

Regarding different environments, I would be very surprised if the latest schemas weren't per environment. However, I'm sure @aaclan-ebi can confirm if you need 100% certainty.

malloryfreeberg:
> Regarding different environments, I would be very surprised if the latest schemas weren't per environment.

Yes, the latest schemas are per-environment. Thanks for clarifying that ingest only allows latest.

justincc:
I also don't think that things are different for analysis ingestion but I'm rather less certain here. @aaclan-ebi will be able to confirm.

samanehsan:
Thanks for the clarification here @justincc and @malloryfreeberg!

malloryfreeberg:
@zperova can we re-assign this PR to @cjvillar to review/merge so he can practice while he's here?

aaclan-ebi:
Hi, just to add, Ingest validates the metadata against the schema specified in the describedBy field. If the metadata json is using an older version of a schema (i.e. describedBy contains an older version of the schema) and the json is valid against that older version, that metadata will be marked as valid. 

@samanehsan , for now, greenbox must ensure that they're sending metadata which uses the latest versions. I'm uncertain if Ingest should enforce that the metadata being submitted should always be the latest. @justincc @simonjupp any thoughts?

justincc:
Might using the non-latest schema for analysis cause problems downstream?

Issue #539, opened 2018-09-10
Values entered as dates only should be valid in date-time typed fields


malloryfreeberg:
**For which schema is a change/update being suggested?**

Affected fields are in the `cell_line.json`, `death.json`, and `specimen_from_organism.json` schemas.

**What should the change/update be?**

The following fields are currently typed to JSON `date-time`. As per the JSON validator, valid values _must_ contain the hh:mm:ss part, but oftentimes the time is not recorded or only a year/year-month is allowed to be supplied. In these cases, a user's value will appear "Invalid", which is not desirable.

1. `cell_line.date_established`: change from JSON type `date-time` to type `date` given that we ask for the "date" the cell line was established, not also the "time".' _If_ we think that the time a cell line was established is important to understand the dataset, we should change the field name to `time_established`. (#821)
1. `death.time_of_death`: only supplying a "date" should be valid. Also, only supplying the year-month or just the year should be valid.
1. `specimen_from_organism.collection_time`: only supplying a "date" should be valid. Also, only supplying the year-month or just the year should be valid.

**What new field(s) need to be changed/added?**

No new fields requested.

**Why is the change requested?**

For an example dataset given to us with `time_of_death` information, only the death year-month was recorded and thus a dummy "00T00:00:00Z" bit had to be added to the end of the date to be valid against the JSON `date-time` format. Expecting users to do this is not appropriate and might not fit the ethical standards for the data. 

`collection_time` might also have some ethical restrictions in cases where a specimen was collected from a dead donor, and the specimen `collection_time` was recorded (thus potentially identifying the donor).

For `date_established`, we want to be sure that we are expecting a date if the word "date" is in the field name and a time if the word "time" is in the field name.

willrockout:
@malloryfreeberg Out of curiosity if these formats are changed to just `Date` with a error be thrown if time is included?

willrockout:
Then do we want to keep this format since it will negate people being able to put time if they have it?

malloryfreeberg:
If a field like `cell_line.date_established` specifically asks for a date, then we should assert that time won't be accepted and ask data contributors to correct it (either in their spreadsheet or, later, directly in the UI) if they entered a time. If a field like `death.time_of_death` could potentially have a time, a user should be able to omit the time and just give a date, for personal identification reasons.

diekhans:
Another approach would have the spreadsheet or submission
program drop the time.

malloryfreeberg:
[...]

PR #962, opened 2019-04-19
Added start and end time to type process. Fixes #742

### LINKED ISSUE #742: Add process_start/end_time to process.json

I would like to request an update to the `process.json` type schema.

I would like to add the new fields - `process_start_time`, `process_end_time` - to this schema to allow data contributors to add when a process began/finished.

1.
* Field name: process_start_time
* Field description: When the process started, in date-time format, yyyy-mm-ddThh:mm:ssZ.
* Required: no
* Example: 2019-01-16T12:00:00Z
* Guideline: Only include if providing the time doesn't make a donor re-identifiable.
2.
* Field name: process_end_time
* Field description: When the process finished, in date-time format, yyyy-mm-ddThh:mm:ssZ.
* Required: no
* Example: 2019-01-16T12:00:00Z
* Guideline: Only include if providing the time doesn't make a donor identifiable.

A data contributor would like to include the start/end times for a process to help identify possible batch effects as their processes were completed months apart. A discussion needs to be had about whether or not to reintroduce these fields with the guideline to only include if they don't make a donor identifiable.

willrockout:
### Release notes
For ```process.json``` schema:

-  Added `start_time`

- Added `end_time`

### Reviews requested

- Need 2 Reviewers to approve because this is a minor update
- Last Reviewer to review/approve, merge changes by 26/04/19


json_schema/type/process/process.json
```diff
@@ -40,6 +40,22 @@
             "$ref": "core/process/process_core.json",
             "user_friendly": "Process core"
         },
+        "start_time": {
```
malloryfreeberg: Since this is a date-time field, should the field names be `start_date_time` and `end_date_time`?

willrockout: Good point! I'm torn on this as the example driving this addition has just a difference in date with no time provided so I'm still on the fence for the mandatory date-time format.

malloryfreeberg: This seems related to the issue https://github.com/HumanCellAtlas/metadata-schema/issues/539 where we need to be able to validate either partial dates, just dates, or date-times. I liked @daniwelter 's idea of having our own regex and implementing `oneOf`: https://github.com/HumanCellAtlas/metadata-schema/issues/539#issuecomment-465535372

Maybe we can do that here?

malloryfreeberg: Updating with `oneOf` to allow either `date-time` or a regex that matches YYYY, YYYY-MM, or YYYY-MM-DD (as per https://github.com/HumanCellAtlas/metadata-schema/issues/966).

json_schema/update_log.csv
```diff
@@ -1 +1,3 @@
 Schema,Change type,Change message,Version,Date
+type/process/process.json,minor,"Added start_date_time to process.json. Fixes #742.",,
```
malloryfreeberg: Remove "`.json`" from the Schema bit

willrockout:
@malloryfreeberg @cjvillar Fixed update log can you review

json_schema/update_log.csv
```diff
@@ -1 +1,3 @@
 Schema,Change type,Change message,Version,Date
+type/process/process,minor,"Added start_date_time to process.json. Fixes #742.",,
+type/process/process,minor,"Added end_date_time to process.json. Fixes #742.",,
```
cjvillar: We should be more conservative with our version increments. Technically these changes are related so we could squeeze them in one line. Also, we should try to keep the update log as accurate as possible, " added start_time"  But I approved and merged.

malloryfreeberg: I agree with @cjvillar , that adding these two related, optional fields should constitute a single minor version increment, not two. And - for consistency in the changelog - I would request that `start_date_time` and `end_date_time` be replaced with the actual names of the new fields.

daniwelter: Butting in here to say +1 on limiting version number increments. Separate lines in the update log should ideally only apply to the same schema is there are different change types, and even then, the highest change type should carry the others.

malloryfreeberg:
@cjvillar you're up to review/merge!

Issue #813, opened 2019-02-12
targets.json should be replaced by probes.json


hewgreen:
**What is the proposed name of the new schema?**

1. I suggest changing the name of the module target.json to probe.json.
2. Rename 'reagent_name' to 'probe_label' this should be required
3. Rename 'molecule_id' to 'target_molecule_label' and add multiple ontologies to support chebi plus proteins and gene products. Ontology code is currently suggested but this is not an ontology field.
4. Remove 'molecule_name'
5. 'channel_id' should become 'channel_label'

**Why is the new schema needed?**

1
There are often multiple probes per target and therefore the granularity of the entity here should be probes rather than target. Users have naturally added multiple targets per row.

2-5
'_id' when used as primary key will be replaced with '_label' in other places in the schema. The primary key will no longer be molecule_id but 'probe_label'

hewgreen:
NB this may be a good opportunity to nest this array in the imaging_protocol.json rather than having this as a module. Waiting for decisions on this from @simonjupp and @diekhans

zperova:
@hewgreen I have further comments/questions.

First, I think we should leave the name of the schema as is because the information we are collecting is about the target, the probe is the implementation, in this case. 

Second I propose to have the following fields:
target label: cFos_ex1
target name: FBJ osteosarcoma oncogene 
target specification: part of exon 1 of cFos
NCBI accession (or ensemble): NM_010234.2
target subcellular structure: nuclear
probe label: cFos_probe1
probe sequence: GATGTTCTCGGGTTTCAACGCCGACTACG
Manufacturer (or in house): in house
assay type: MERFISH
channel label: 1
fluorophore (if non-multiplexed): 
multiplexed?: yes

What do you think?

hewgreen:
Your first point is not true as per point 1. As these experiments use multiple probes per target, contributors hacked the sheet to reflect their assay. They were forced to provide probes as entities rather than targets. When the mapping probe:target is 1:1 either name works but as the majority will be multiple probes per target this is required. One entity per row is consistent with the rest of the schema. (A second implementation would be to have another nested array so that we can provide lists of probes per target but then we are nesting too deeply.)

Regarding your list of attributes I don't see much difference. Although I wonder if we could combine `target subcellular structure`, `target label`, `target name` and `target specification` to a single ontology field. This would add a tremendous amount of clarity to the metadata. I think hiding the flexibility and complexity is a good idea here.

zperova:
[...]

PR #978, opened 2019-05-03
Delete deprecated target.json schema. Fixes#813.

zperova:
### Release notes

Deleted deprecated target.json module.
Probes.json replaced targets.json as per #813 and #886 
 

### Reviews requested

- Needs @daniwelter to approve because this is technically not a version change


zperova:
This is blocked by #981  - Travis won't pass until it is merged

zperova:
This PR is now combines with #981

PR #981, opened 2019-05-03
Removed target ref in imaging_protocol.json. Fixes#980.

zperova:
### Release notes

For imaging_protocol.json schema:
- replaced "target" with "probe" in all relevant fields. 
 
For deprecated target.json:
delete this schema

### Reviews requested

- Need 2 Reviewers to approve because this is a bug fix during metadata freeze
- Last Reviewer to review/approve, merge changes by May 13


daniwelter:
LGTM! @malloryfreeberg, ready for you when you are.

malloryfreeberg:
@zperova @daniwelter I approved but didn't merge this PR yet. I was wondering what we wanted to decide about merging major changes to develop while we are in Q2 freeze. If we merge major changes to dev, we won't be able to merge patch or minor changes (which shouldn't break software) during the freeze. I think the last time we did a freeze, we merged everything to dev but that prevented us from pushing patch and minor changes to prod (the patch changes were useful for the portal and spreadsheet generation). Thoughts?

zperova:
@malloryfreeberg this can wait until the are out of freeze. I didn't expect to make a major change, but once I found the mistake it turned out a major change to fix it. I think we should leave ourselves a way to push the minor and patch updates during the freeze.

daniwelter:
No objections in principle to holding off on PR merges for major changes until after the Q2 freeze given the version number chaos that would result from patch changes

malloryfreeberg:
If this is a bug fix, then it should be a patch and should be OK to merge.

malloryfreeberg:
I think this should be a patch/bug change and not a major change. Although the change alone is "major", the change is actually fixing something that would have been "broken" or at least incorrect otherwise. If we make this a bug fix, then I think there is an argument to merge it in now during the Q2 metadata freeze.

zperova:
@malloryfreeberg @daniwelter great I am happy to make it a bug fix. Could someone please merge it?

PR #1017, opened 2019-06-05
Changed travis conf file for multi-stage build. Fixes #1009

### LINKED ISSUE #1009: Implement Travis multi-stage build for the schema linter.
**Description**

 

As a DCP member who updates metadata schemas, I would like all PRs to pass the linting test before any other tests using the Travis multi-stage build.

From @rexwangcc : If the linting test is fast and you want all runs to pass the linting test before any other tests, it might be good to use the Travis multi-stage build. We used it this way [here](https://github.com/HumanCellAtlas/pipeline-tools/blob/master/.travis.yml) which results in: https://travis-ci.org/HumanCellAtlas/pipeline-tools/builds/531931757

**Acceptance Criteria**

  
- [x] Travis rung multi-stage builds and reports results like [this](https://travis-ci.org/HumanCellAtlas/pipeline-tools/builds/531931757)

ESapenaVentura:
### Release notes

No schema was changed. Changes to the configuration file (`.travis.yml`) were as follows:

- Separated schema linting from schema testing.
- Changed order of jobs. Since schema linting is lighter, it will be performed first.
- Added a channel for slack notifications (`metadata-ci-test`, will change name and make public once we know it's working properly). 
 

### Reviews requested

- Need 1 Reviewers to approve because this is a script update
- Last Reviewer to review/approve, merge changes by 12/06/2019


.travis.yml
```diff
@@ -1,21 +1,22 @@
...
+- '3.6'
+jobs:
+  include:
+  - stage: Schema linter
+    script:
+       - cd src
+       - python schema_linter.py
+  - stage: Schema tests
+    install:
+       - cd ../tests
```
malloryfreeberg: I never liked have this `cd ../` part here. It means that if someone rearranged the tests, the relative paths would be broken. Could we maybe add `cd ../` at the end of each script, so that each script always starts at the root dir? That way, if someone wanted to rearrange or add tests, each script would be starting from the same root dir.

ESapenaVentura:
@HumanCellAtlas/wranglers How do we want the notifications for slack? Here we have the options: https://docs.travis-ci.com/user/notifications/ 

Which basically resumes in: notify on success or on fail when: build status changes; always; never. It defaults to always, but that might be way too many notifications for a repo that changes that much. Maybe we could get it to on_success: change and on_fail: always?

Thoughts?

malloryfreeberg:
> on_fail: always

I don't know if I want a notification every time the build fails :) In this PR alone that would have been 3 notifications!

ESapenaVentura:
> > on_fail: always
> 
> I don't know if I want a notification every time the build fails :) In this PR alone that would have been 3 notifications!

It's settled, then! both on change!

malloryfreeberg:
> both on change!

So that means every time the build goes from green > red and red > green?

ESapenaVentura:
>
> So that means every time the build goes from green > red and red > green?

Yes!

PR #1077, opened 2019-07-24
Adding two new fields representing the major and minor version of the schema in each of the type metadata schemas. Fixes #1076.

### LINKED ISSUE #1076: Add two optional fields to metadata schemas
In order to enable rapid integration of data adhering to changed metadata schemas and in accordance with the design detailed here, https://docs.google.com/document/d/1BInKossmrlpna5MV_s4MSR0Bll1wnD-3OVw-9nCLdWg/edit?ts=5d389d03#, create two additional fields in the metadata schemas: 

"schema_major_version": {
      "description": "The major version number of the schema.",
      "type": "integer",
      "pattern": "^[1-9]\d*$",
      "example": "4"
    },
"schema_minor_version": {
      "description": "The minor version number of the schema.",
      "type": "integer",
      "pattern": "^\d*$",
      "example": "6"
    }

maniarathi:
### Release notes

~For all Type schemas:~
For provenance.json schema:
- Added new fields schema_major_version and schema_minor_version which are representing the major and minor schema version respectively. These are optional fields for now which will eventually be turned into required fields once all data that is today in the DCP has been back-populated.
 

### Reviews requested

- Need 2 Reviewers to approve because this is a minor update
- Last Reviewer to review/approve, merge changes by Monday, July 29th.


simonjupp:
I don’t think this should go into the schema. We’re asking programmatic contributors to provide redundant information. This should live in the json schemas themselves and components should get it by dereferencing the schema. 

An alternative might be for ingest to add this into the provenance block.

diekhans:
> I don’t think this should go into the schema.

@simonjupp  do you mean in the instances,

simonjupp:
Yes

simonjupp:
And I’d love to see schema_type go while we’re at it.

maniarathi:
Trying to understand the feedback a bit better: why include describedBy, schema_version, and schema_type at the top level if they are already represented in the provenance block? Am I correct in my understanding that the provenance block is entirely generated by Ingest?

simonjupp:
None of these aren’t currently in the provenance block and yes, the provenance block is reserved for anything added by ingest.

simonjupp:
My suggestion for now is to keep describe_by at the top level. Then move schema_type and major/minor version into the provenance block and ask ingest to insert this information when a document it is created.

maniarathi:
Hmm, are we talking about the same thing? I was referring to https://github.com/HumanCellAtlas/metadata-schema/blob/master/json_schema/system/provenance.json#L13.

simonjupp:
I forgot that we have a described_by and a schema_version in every single schema! This is not necessary and some legacy bit of schema that never got cleaned up properly. 

Only the top level schemas (under the type) directory should have a described_by properly. We should remove described_by from all system, module and core schemas as it is redundant and not something we would ever expect a contributor to use.

justincc:
Yes, on reflection I think ingest will need to parse describedBy and insert minor/major versions. Otherwise we're requiring submitters to fill in more and redundant information and it opens up the possibility of inconsistency.

I hoped we could avoid this in the first instance to reduce the need for inter-component coordination and a longer implementation time. But that position feels unsustainable.

tburdett:
I agree with @simonjupp; programmatic submitters and ingest tooling shouldn't be providing redundant information. However, I also see the motivation from @maniarathi - we need a pragmatic solution to allow consumers of the JSON to index the version of a schema and provide searches over it. I see two options:

1. Providers of metadata must provide redundant information
2. Services offering metadata presentation must dereference the `describedBy` link to collect additional content (major, minor versions) from the schema

Doing 1. will require the ingestion service to implement the functionality in 2. for all non-programmatic submitters (as @justincc comments above) **and also** will require pipelines to implement 2. for analysis submissions.

If levels of technical debt in other downstream components are so high that this is the expedient approach, this might be the right option to choose, but I think this is **only** a good solution if we have a plan to refactor out this workaround in future. Long-term, asking data providers to supply information that is system-required (and redundant) seems like a recipe for disaster to me.

simonjupp:
I think the quickest solution with the minimal disruption is to put the major and minor version into `provenance.json` ( then have ingest set these via deref'ing or querying the schemas API when  a doc is received). 

This 1) avoids redundancy  2) avoids adding something that we plan to remove in the future and 3) avoids the need to makes any changes in the existing importer or analysis submission tools.

malloryfreeberg:
Ok, I think I agree that for the sake of moving forward with the metadata migration plan and moving towards the optimal state where `describedBy` is also in `provenance.json`, we should stick these 2 new fields in `provenance.json` as optional fields. I can do the change and then request the reviews.

maniarathi:
Thanks for all the input and the discussion! I'm totally happy with this change and glad to get more clarification on the nuances of metadata.

PR #1089, opened 2019-08-02
Inside a character class, you don't need "|" (no bug filed)

mccalluc:
(... unless you actually want to match "|")

### Release notes

For biomaterial_core.json schema:
- Fix regex syntax mistake
 
### Reviews requested

- Need 1 Reviewers to approve because this is a patch update
- Last Reviewer to review/approve, merge changes by: no-hurry


ESapenaVentura:
I'll take care of it. Thanks for the observation.

ESapenaVentura:
Hi @mccalluc! Thanks for taking your time to collaborate to make our schemas better. Due to how we work in the metadata repository, we prefer the changes to be pushed in from a branch rather than a fork. Also, we've seen that you have another request regarding a regex on the same schema (#1090).

If you don't mind, I'll proceed to close this pull request, and create another one that fulfill both requested changes. Thanks!

PR #1176, opened 2019-11-25
Release from staging to production - 2019/11/26

willrockout:
### Release notes
No major changes


ESapenaVentura:
@willrockout I have no idea what this commits are. I can only think that somehow I did something wrong when merging the snyk's PR to different environments, but no file seems to have changed in these commits :/

PR #1189, opened 2019-11-29
Fix insdc accessions Fixes #1167

### LINKED ISSUE #1167: Fix INSDC accessions for project and study in the project.json schema

I would like to request an update to the `project.json` schema.

I would like to fix the following two fields: 

`insdc_project_accessions` and `insdc_study_accessions`

For the field `insdc_project_accessions` the following should be changed to:

"pattern": "^PRJ[E|N|D][a-zA-Z][0-9]+$"
"example": "PRJNA000000"
"guidelines": "Enter accession if project has been archived. Accession can be from the DDBJ, NCBI, or EMBL-EBI and must start with PRJD, PRJN, or PRJE, respectively."

For the field `insdc_study_accessions` the following should be changed to:

"pattern": "^[D|E|S]RP[0-9]+$"
"example": "SRP000000"
"guidelines": "Enter accession if study has been archived. Accession can be from the DDBJ, NCBI, or EMBL-EBI and must start with DRP, SRP, or ERP, respectively."

This information is incorrect as it is currently mixed up. 

This is a major change as breaks compatibility of data currently in the system.

mshadbolt:
### Release notes

For type/project/project.json schema:
- Fixed regex, example and description for insdc_study_accessions, insdc_project_accessions, array_express_accessions, biostudies_accessions, geo_series_accessions 

### Reviews requested

- Need 2 Reviewers to approve because this is a major update. It makes data currently submitted invalid against the latest schema due to the insdc study/project regex swap
- Last Reviewer to review/approve, merge changes by Thursday, 5th December


ESapenaVentura:
Major metadata schema update, pinging @maniarathi @aaclan-ebi @MDunitz @NoopDog @brianraymor @jahilton @jychien @jlzamanian @samanehsan @hannes-ucsc. If no one objects, this will be merged on the following 5 working days.

json_schema/type/project/project.json
```diff
@@ -71,55 +71,55 @@
             "type": "array",
             "items": {
                 "type": "string",
-                "pattern": "^[D|E|S]RP[0-9]+$"
+                "pattern": "^PRJ[E|N|D][a-zA-Z]\\d+$"
```
ESapenaVentura: `[a-zA-Z]` Does the accession number accept lowercase letters?

mshadbolt: Not really sure, I just copied from what it was before...

mshadbolt: according to this it should just be upper case I think https://ena-docs.readthedocs.io/en/latest/submit/general-guide/accessions.html

mshadbolt: I removed the lower case letters

json_schema/type/project/project.json
```diff
@@ -71,55 +71,55 @@
...
-            "guidelines": "Enter accession if project has been archived. Accession must start with E-."
+            "guidelines": "Enter accession if project has been archived in ArrayExpress. Accession must start with E-."
         },
         "insdc_study_accessions": {
             "description": "An International Nucleotide Sequence Database Collaboration (INSDC) study accession.",
             "type": "array",
             "items": {
                 "type": "string",
-                "pattern": "^PRJ[E|N|D][a-zA-Z][0-9]+$"
+                "pattern": "^[D|E|S]RP\\d+$"
```
hannes-ucsc: This regex matches `|RP123`. So either, 

```suggestion
                "pattern": "^[DES]RP\\d+$"
```

or 

```suggestion
                "pattern": "^(D|E|S)RP\\d+$"
```

But the former is far more common when the alternatives are individual characters. `|` is typically used when the alternatives are complex regexes in their own right.

json_schema/type/project/project.json
```diff
@@ -71,55 +71,55 @@
             "type": "array",
             "items": {
                 "type": "string",
-                "pattern": "^[D|E|S]RP[0-9]+$"
+                "pattern": "^PRJ[E|N|D][A-Z]\\d+$"
```
hannes-ucsc: Same here.

mshadbolt: Thanks @hannes-ucsc ! I have fixed the regex patterns

mshadbolt:
Maybe they were copy pasted from here: https://ena-docs.readthedocs.io/en/latest/submit/general-guide/accessions.html originally, but these are not in the correct regex format. I have seen a few times that out regex has pipes in it which it shouldn't so we should keep an eye out and correct if there are any more instances.

mshadbolt:
I think I have addressed all reviewers comments so if one of you is able to merge that would be grand.

ESapenaVentura:
Hi @zperova the quotes are not necessary. Ran the `release_prepare.py` locally and saw that the comment was updated in the changelog correctly, so it is ready to go. I have pushed the changes and I'm ready to merge if you think it's ok!

PR #1284, opened 2020-05-28
Made changes to links.json schema. Fixes #1285

### LINKED ISSUE #1275: Remove `input_bundles` from `analysis_process`
As specified in the [DCP2 MVP Schema doc](https://docs.google.com/document/d/1NsibP8g-NeLnksxlcBWQsSj5Zg_uimCDaAF1qB_qkjg/edit#heading=h.eeyjdjt197l9)

`type/process/analysis_process.json`

remove the field `input_bundles` ~and replace it with `input_files`~

* ~Field name: input_files~
* ~Field description: A list of references to the metadata entities that describe the individual input data files~
* ~Field type: array~
* ~item type: string~
* ~Required: yes~
* ~user_friendly: Input files~
* ~Examples: b816d2d6-5f10-4447-4194-3d0a804454d6~
* ~CV or enum: no~

### LINKED ISSUE #1285: modify links.json to allow for explicit supplementary file linking

`system/links.json`

As part of work for DCP2 MVP we need a way of explicitly linking supplemental files to a project. The suggestion is to modify the links.json file to include both `experimental_graph_link` and `supplementary_file_link`.

Supplementary files should be able to be linked to any metadata entity whether that is the project, protocol or biomaterial.

change the current `link` definition to `experimental_graph_link`, i.e. line 83-91 to read

```
"experimental_graph_link": {
      "type": "object",
      "required": [
        "process_type",
        "process_id",
        "inputs",
        "outputs",
        "protocols"
      ]
...
```

add a new type of `link` object, `supplementary_file_link`

```
"supplementary_file_link": {
      "type": "object",
      "required": [
        "inputs",
        "file_entity_id"
      ],
      "properties": {
        "inputs": {
          "description": "An array of inputs for this link",
          "type": "array",
          "items": {
            "type": "object",
            "$ref": "#/definitions/input"
          },
          "file_entity_id": {
            "description": "A link to a supplementary file uuid"
            "type": "string"
        }
```

modify the "links" definition to allow objects of both the new link types e.g.

```
"links": {
      "description": "An array of linking objects.",
      "type": "array",
      "items": {
        "type": "object",
        oneOf:
          - "$ref": "#/definitions/experimental_graph_link"
          - "$ref": "#/definitions/supplementary_file_link"
      }
```
tag @ESapenaVentura @rolando-ebi

ESapenaVentura:
### Release notes

For `system/links.json`:
- Added definitions and non-required `supplementary_file_link` property within `links` to allow for supplementary file linking.
 

### Reviews requested

- This update will be applied as a hotfix once it's approved

### DIFF
--- a/json_schema/property_migrations.json
+++ b/json_schema/property_migrations.json
@@ -1,15 +1,6 @@
 {
     "migrations": [
         {
-            "source_schema": "analysis_process",
-            "property": "input_bundles",
-            "target_schema": "analysis_process",
-            "replaced_by": "input_files",
-            "effective_from": "12.0.0",
-            "reason": "Field clarification",
-            "type": "renamed property"
-        },
-        {
             "source_schema": "cell_suspension",
             "property": "total_estimated_cells",
             "target_schema": "cell_suspension",
--- a/json_schema/property_migrations.json
+++ b/json_schema/property_migrations.json
@@ -1,6 +1,15 @@
 {
     "migrations": [
         {
+            "source_schema": "analysis_process",
+            "property": "input_bundles",
+            "target_schema": "analysis_process",
+            "replaced_by": "input_files",
+            "effective_from": "12.0.0",
+            "reason": "Field clarification",
+            "type": "renamed property"
+        },
+        {
             "source_schema": "cell_suspension",
             "property": "total_estimated_cells",
             "target_schema": "cell_suspension",
--- a/json_schema/property_migrations.json
+++ b/json_schema/property_migrations.json
@@ -1,6 +1,27 @@
 {
     "migrations": [
         {
+            "source_schema": "analysis_process",
+            "property": "input_bundles",
+            "effective_from": "12.0.0",
+            "reason": "Field not used in DCP2",
+            "type": "deprecated property"
+        },
+        {
+            "source_schema": "analysis_process",
+            "property": "reference_files",
+            "effective_from": "12.0.0",
+            "reason": "Linking reference files to analysis process",
+            "type": "new required property"
+        },
+        {
+            "source_schema": "analysis_process",
+            "property": "reference_bundle",
+            "effective_from": "12.0.0",
+            "reason": "Field not used in DCP2",
+            "type": "deprecated property"
+        },
+        {
             "source_schema": "cell_suspension",
             "property": "total_estimated_cells",
             "target_schema": "cell_suspension",
--- a/json_schema/system/links.json
+++ b/json_schema/system/links.json
@@ -80,14 +80,15 @@
                 }
             }
         },
-        "experiment_link": {
+        "process_link": {
             "type": "object",
             "required": [
                 "process_type",
                 "process_id",
                 "inputs",
                 "outputs",
-                "protocols"
+                "protocols",
+                "link_type"
             ],
             "properties": {
                 "process_type": {
@@ -126,6 +127,13 @@
                         "type": "object",
                         "$ref": "#/definitions/protocol"
                     }
+                },
+                "link_type": {
+                    "description": "The type of this link.",
+                    "type": "string",
+                    "enum": [
+                        "process_link"
+                    ]
                 }
             }
         },
@@ -174,7 +182,8 @@
             "type": "object",
             "required": [
                 "entity",
-                "supplementary_files"
+                "supplementary_files",
+                "link_type"
             ],
             "properties": {
                 "entity": {
@@ -189,6 +198,13 @@
                         "type": "object",
                         "$ref": "#/definitions/supplementary_file"
                     }
+                },
+                "link_type": {
+                    "description": "The type of this link.",
+                    "type": "string",
+                    "enum": [
+                        "supplementary_file_link"
+                    ]
                 }
             }
         }
--- a/json_schema/system/links.json
+++ b/json_schema/system/links.json
@@ -1,230 +1,230 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
-  "description": "A schema for describing the links between entities in an experiment graph.",
-  "additionalProperties": false,
-  "required": [
-    "describedBy",
-    "schema_type",
-    "links"
-  ],
-  "title": "Links",
-  "name": "links",
-  "type": "object",
-  "definitions": {
-    "input": {
-      "type": "object",
-      "description": "An input to a process",
-      "required": [
-        "input_type",
-        "input_id"
-      ],
-      "properties": {
-        "input_type": {
-          "description": "The concrete type of the input entity.",
-          "example": "cell_suspension",
-          "user_friendly": "Input type",
-          "type": "string"
-        },
-        "input_id": {
-          "description": "UUID of the input entity.",
-          "example": "111f222a-333b-444c-555d-eee63ddfccc7",
-          "user_friendly": "Input entity UUID",
-          "type": "string",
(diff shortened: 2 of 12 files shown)


ESapenaVentura:
This change will be hotfixed for every environment as it's necessary for the DCP2 MVP

ESapenaVentura:
@hannes-ucsc We've modified the PR to get both the requested change for the `analysis_process` schema and further changes to `links.json` schema to allow linking of supplementary files to any entity (Although for now all supplementary files will be linked to the project).

Please let us know if you'd like to suggest changes

json_schema/system/links.json
```diff
@@ -80,7 +80,7 @@
         }
       }
     },
-    "link": {
+    "experiment_link": {
```
hannes-ucsc, line 83: I would use

```suggestion
    "process_link": {
```

"experiment" may be overly restrictive. After all, each link represents a process.

json_schema/system/links.json
```diff
@@ -128,33 +128,103 @@
...
+            "enum": [
+              "links"
+            ]
+          },
+          "links": {
+            "description": "An array of linking objects.",
+            "type": "array",
+            "items": {
+              "type": "object",
+              "oneOf": [
```
hannes-ucsc, line 219: Since you are introducing polymorphism, you should consider adding a discriminator property to make it easier for metadata consumers to distinguish the types. Without a discriminator, consumers would need to apply either a heuristic (probing for certain distinguishing properties) or a full schema validation in order to determine which subtype of link they are dealing with. This could be done by adding a `link_type` property in each of the subschemas and pin it to either `process_link` or `supplementary_link` via a singleton enum.

json_schema/system/links.json
```diff
@@ -128,33 +128,103 @@
...
+        }
+      }
     },
-    "links": {
-      "description": "An array of linking objects.",
-      "type": "array",
-      "items": {
-        "type": "object",
-        "$ref": "#/definitions/link"
+    "supplementary_file_link": {
```
hannes-ucsc, line 173: Could shorten this to

```suggestion
    "supplementary_link": {
```

json_schema/system/links.json
```diff
@@ -128,33 +128,103 @@
...
+      "properties": {
+         "entity_id": {
+            "description": "ID of this entity",
+            "example": "111f222a-333b-444c-555d-eee63ddfccc7",
+             "type": "string",
+             "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+         },
+        "entity_type": {
+          "description": "The concrete type of this entity",
+          "type": "string"
```
hannes-ucsc, line 169: I think this should be an enum or a pattern and explicitly list the concrete types that we, the DCP, support at this time. This, by the way, applies to all of the `…_type` properties. You could also implement the restrictions in a separate graph validator but not every consumer has access to that. JSONSchema is a standard and has off-the-shelves implementations, so we should use it to express graph constraints whenever possible. Our transitioning to concrete types in the previous PR for `links.json` enables us to do so.

hannes-ucsc:
A note on process, if I may. You may want to factor out the `input_file` changes into a separate PR. These two changes have little to do with each other. One of them is less intrusive than the other and risks being taken hostage by a the discussion on that other change.

ESapenaVentura:
Based on the comments by @hannes-ucsc above, we are going to keep this PR with only the `links.json` schema changes for further discussion and move the changes to `analysis_process.json` schema changes to a new PR.

json_schema/system/links.json
ESapenaVentura: Wouldn't this suggest a link that is supplementary, and not a link to a supplementary file?

rolando-ebi:
@hannes-ucsc 

> Generally this goes in the right direction. But for DCP/2 MVP, we need to think how we would rewrite the existing `links.json` in v1 to this new schema. The only thing we know that a supplementary file (SF) is associated with a bundle and indirectly to a project. 

Yeah this would involve having to try to parse this info out of a v1 bundle. All v1 bundles should have a project. For any bundle that contains supplementary files (many do not), we can assume that they were all supplementary files for the project since ingest only allows supplementary files for project entities. So each v1 bundle would contain, at most, only one `supplementary_file_link`:

```
"links": [
  ...
  ...
  {
    "entity": {
        "entity_type": "project"
        "entity_id": "<project uuid>"
    },
    "supplementary_files": [
       <each file-metadata in the bundle with schema_type `supplementary_file`>
    ]
  }
]

```

json_schema/system/links.json
hannes-ucsc: It would and I think that's ok. I can't imagine a scenario in which a link to a supplementary file is not supplementary itself. MW says "supplementary: available to supply something extra when needed" so I think supplementary-ness is transitive, so to speak.

However, this is a cosmetic issue for me. Either name works for me.

hannes-ucsc:
@rolando-ebi I completely agree. @danielsotirhos would be implementing that aspect of rewriting links.json. If you accept my suggestion of restricting `links[*].entity_type` we could initially just permit `project` as a value for that property and open it up as needed.

json_schema/property_migrations.json
```diff
@@ -1,5 +1,14 @@
 {
     "migrations": [
+        {
```
mshadbolt: should this still be in this pull request?

json_schema/system/links.json
mshadbolt: process link seems ok to me

json_schema/system/links.json
mshadbolt: I'd also prefer to keep `_file_` as it could be confusing as we have the concept of 'supplementary links' to link to websites/github etc from the project

json_schema/property_migrations.json
ESapenaVentura: It says to me that this is outdated and looking at the files this shouldn't appear anymore

rolando-ebi:
@hannes-ucsc 

> If you accept my suggestion of restricting `links[*].entity_type` we could initially just permit `project` as a value for that property and open it up as needed later.

Yep, I think the constraint on the allowable entity_type values is a good idea. The wranglers would have to maintain this enum as the set of concrete entities change. Should be fairly straightforward to write a unit test to ensure none are missed. For now, will make it so that `project` is the only allowable entity_type. Note as well that `supplementary_file` is similarly the only permitted "file_type" in supplementary file links

json_schema/system/links.json
hannes-ucsc: Agreed.

ESapenaVentura:
@kbergin Pinging you as this PR might be of interest to the analysis_team. Who should review this from your team?

Also PR #1286 might be of interest for your team to review as well

ESapenaVentura:
@hannes-ucsc Rolando pushed the changes that were requested, does this PR sound reasonable?

json_schema/system/links.json
```diff
@@ -1,160 +1,249 @@
...
+                "supplementary_files",
+                "link_type"
+            ],
+            "properties": {
+                "entity": {
+                    "description": "An entity that refers to this supplementary file",
+                    "type": "object",
+                    "$ref": "#/definitions/entity"
+                },
+                "supplementary_files": {
```
hannes-ucsc, line 197:
```suggestion
                "files": {
```

Just a suggestion, to reduce the verbosity. Within a supplementary file link it should be clear that a linked file is a supplementary file.

ESapenaVentura: LGTM!

PR #1286, opened 2020-06-01
Changed reference_bundle to files and removed input_bundles. Fixes #1275 and #1290

### LINKED ISSUE #1275: Remove `input_bundles` from `analysis_process`

ESapenaVentura:
### Release notes

For `type/process/analysis/analysis_process.json` schema:
- Removed `input_bundles`
- Changed `reference_bundle` to `reference_files` and updated the field's properties.
- Changed required fields to reflect the new property name and removal of input_bundles.

 

### Reviews requested

- This update will be applied to all environments once it's approved.

### DIFF
--- a/json_schema/property_migrations.json
+++ b/json_schema/property_migrations.json
@@ -9,12 +9,17 @@
         },
         {
             "source_schema": "analysis_process",
-            "property": "reference_bundle",
-            "target_schema": "analysis_process",
-            "replaced_by": "reference_files",
+            "property": "reference_files",
             "effective_from": "12.0.0",
-            "reason": "Field re-purposed for DCP2",
-            "type": "renamed property"
+            "reason": "Linking reference files to analysis process",
+            "type": "new required property"
+        },
+        {
+            "source_schema": "analysis_process",
+            "property": "reference_bundle",
+            "effective_from": "12.0.0",
+            "reason": "Field not used in DCP2",
+            "type": "deprecated property"
         },
         {
             "source_schema": "cell_suspension",
--- a/json_schema/property_migrations.json
+++ b/json_schema/property_migrations.json
@@ -1,6 +1,24 @@
 {
     "migrations": [
         {
+            "source_schema": "analysis_process",
+            "property": "input_bundles",
+            "target_schema": "analysis_process",
+            "replaced_by": "input_files",
+            "effective_from": "12.0.0",
+            "reason": "Field clarification",
+            "type": "renamed property"
+        },
+        {
+            "source_schema": "analysis_process",
+            "property": "reference_bundle",
+            "target_schema": "analysis_process",
+            "replaced_by": "reference_file",
+            "effective_from": "12.0.0",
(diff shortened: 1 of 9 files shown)


json_schema/type/process/analysis/analysis_process.json
```diff
@@ -216,11 +217,12 @@
             "user_friendly": "Analysis run type",
             "example": "Should be one of: run, or copy-forward."
         },
-        "reference_bundle": {
-            "description": "UUID of the bundle containing the reference used in running the pipeline.",
+        "reference_file": {
```
hannes-ucsc, line 220: I think we should account for the possibility of multiple reference files, if they're split by chromosome, for example. A bundle can contain multiple files so that's covered in the current schema.

ESapenaVentura:
As we discussed yesterday, I've made the changes to the PR to reflect the specs in the DCP2 design document.

Asking for a re-revision

json_schema/property_migrations.json
```diff
@@ -1,5 +1,21 @@
...
+            "type": "deprecated property"
+        },
+        {
+            "source_schema": "analysis_process",
+            "property": "reference_bundle",
+            "target_schema": "analysis_process",
+            "replaced_by": "reference_files",
+            "effective_from": "12.0.0",
+            "reason": "Field re-purposed for DCP2",
+            "type": "renamed property"
```
hannes-ucsc, line 17: This can't be modelled as a rename, I think. This should be modelled as a deprecation and an addition. A property rename implies that the value stays the same which isn't the case here.

json_schema/type/process/analysis/analysis_process.json
```diff
@@ -216,11 +206,15 @@
             "user_friendly": "Analysis run type",
             "example": "Should be one of: run, or copy-forward."
         },
-        "reference_bundle": {
-            "description": "UUID of the bundle containing the reference used in running the pipeline.",
-            "type": "string",
-            "user_friendly": "Reference bundle",
-            "example": "b816d2d6-5f10-4447-4194-3d0a804454d6"
+        "reference_files": {
+            "description": "UUID of the file containing the reference used in running the pipeline.",
```
hannes-ucsc, line 210:
```suggestion
            "description": "UUID of the file entities that contain the reference genome used in running the pipeline.",
```

PR #1300, opened 2020-06-09
Added new optional properties for file integrity check. Fixes #1289.

### LINKED ISSUE #1289: Add checksums and content type to file_core
https://docs.google.com/document/d/1NsibP8g-NeLnksxlcBWQsSj5Zg_uimCDaAF1qB_qkjg/edit#heading=h.eeyjdjt197l9

TODO: fill this out

*e.g.*: I would like to request an update to the `cell_line.json` schema.

*e.g.*: I would like to add a new field - `random_integer` - to this schema to allow data contributors to assign a random integer to the cell line.

This update constitutes a \<major/minor/patch> change to the schema(s) it affects.

*e.g.*:
* Field name: random_integer
* Field description: A random integer.
* Field type: integer
* Required: no
* Examples: 42
* CV or enum: no

*e.g.*: Data contributors need to be able to assign a random integer to a cell line in order to keep track of it.

ESapenaVentura:
### Release notes

For core/file/file_core.json schema:
- Added `content_type` property
- Added `size` property
- Added `sha256` property
- Added `crc32c` property
- Added `sha1` property
- Added `s3_etag` property
 

### Reviews requested

- Need 2 Reviewers (But 3 will review due to circumstances) to approve because this is a minor update
- This hotfix will be merged by me once the PR has been reviewed.

### DIFF
--- a/docs/jsonBrowser/core.md
+++ b/docs/jsonBrowser/core.md
@@ -18,6 +18,12 @@ file_name | The name of the file. | string | yes |  | File name |  | R1.fastq.gz
 format | The format of the file. | string | yes |  | File format |  | fastq.gz; tif
 content_description | General description of the contents of the file. | array | no | [See module  file_content_ontology](module.md/#file_content_ontology) | Content description |  | 
 checksum | MD5 checksum of the file. | string | no |  | Checksum |  | e09a986c2e630130b1849d4bf9a94c06
+content_type | An appropriate MIME type for this file. | string | no |  | Content type |  | application/gzip
+size | The size of the file in bytes. | number | no |  | Size |  | 2147483648
+sha256 | The sha256 hash of the file. | string | no |  | Sha256 |  | 12998c017066eb0d2a70b94e6ed3192985855ce390f321bbdb832022888bd251
+crc32c | The CRC-32C code generated for this file. | string | no |  | Crc32c |  | 3FBA199E
+sha1 | The sha1 hash of the file. | string | no |  | Sha1 |  | 6e71b3cac15d32fe2d36c270887df9479c25c640
+s3_etag | The AWS S3 ETag for this file | string | no |  | s3 ETag |  | 
 
 ## Protocol core<a name='Protocol core'></a>
 _Information about an intended protocol that was followed in the project._
--- a/json_schema/core/file/file_core.json
+++ b/json_schema/core/file/file_core.json
@@ -48,7 +48,47 @@
             "description": "MD5 checksum of the file.",
             "type": "string",
             "user_friendly": "Checksum",
-            "example": "e09a986c2e630130b1849d4bf9a94c06"
-	    }
+            "example": "e09a986c2e630130b1849d4bf9a94c06",
+            "pattern": "^[a-f0-9]{32}$"
+	    },
+        "content_type": {
+            "description": "An appropriate MIME type for this file.",
+            "type": "string",
+            "user_friendly": "Content type",
+            "example": "application/gzip"
+        },
+        "size": {
+            "description": "The size of the file in bytes.",
+            "type": "number",
+            "minimum": 0,
+            "user_friendly": "Size",
+            "example": 2147483648
+        },
+        "sha256": {
+            "description": "The sha256 hash of the file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{64}",
+            "user_friendly": "Sha256",
+            "example": "12998c017066eb0d2a70b94e6ed3192985855ce390f321bbdb832022888bd251"
+        },
+        "crc32c": {
+            "description": "The CRC-32C code generated for this file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{8}",
+            "user_friendly": "Crc32c",
+            "example": "3FBA199E"
+        },
+        "sha1": {
+            "description": "The sha1 hash of the file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{40}",
+            "user_friendly": "Sha1",
+            "example": "6e71b3cac15d32fe2d36c270887df9479c25c640"
+        },
+        "s3_etag" : {
+            "description": "The AWS S3 ETag for this file",
+            "user_friendly": "s3 ETag",
+            "type": "string"
+        }
     }
 }
--- a/json_schema/update_log.csv
+++ b/json_schema/update_log.csv
@@ -1,2 +1,2 @@
 Schema,Change type,Change message,Version,Date
-
+core/file/file_core,minor,"Added new optional properties for file integrity check. Fixes #1289.",,


mshadbolt:
Would it be worth adding anything to the description to specify that these are ingest system generated rather than user filled fields?

ESapenaVentura:
@mshadbolt I like the idea but in the description feels a little bit off. What about in `guidelines`?

mshadbolt:
I'm not sure what you mean by guidelines? do we specify that in our schema? Or is it in some other document?

ESapenaVentura:
Sorry, my bad for not putting an example.

For some fields, we have the "guidelines" which are special indications for the field.

An example [here](https://github.com/HumanCellAtlas/metadata-schema/blob/master/json_schema/type/biomaterial/donor_organism.json#L91,L97)

mshadbolt:
oh right, I always thought of those as guidelines for a user so don't know whether that is the best place either

lauraclarke:
The goal of the guidelines field was to enable the writing of more instructional text as well as giving examples in the example section as you don't always want to frame it the same way

I don't think it matters where you put it but flagging these are system supplied in the guidance does seem like a good plan

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+        },
+        "size": {
+            "description": "The size of the file in bytes.",
+            "type": "number",
+            "minimum": 0,
+            "user_friendly": "Size",
+            "example": 2147483648
+        },
+        "sha256": {
+            "description": "The sha256 hash of the file.",
```
hannes-ucsc, line 68:
```suggestion
            "description": "The SHA-256 hash of the file.",
```

That's the name used by the official NSA standard document.

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{64}",
+            "user_friendly": "Sha256",
```
hannes-ucsc, line 71:
```suggestion
            "user_friendly": "SHA-256",
```

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+        },
+        "crc32c": {
+            "description": "The CRC-32C code generated for this file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{8}",
+            "user_friendly": "Crc32c",
+            "example": "3FBA199E"
+        },
+        "sha1": {
+            "description": "The sha1 hash of the file.",
```
hannes-ucsc, line 82:
```suggestion
            "description": "The SHA-1 hash of the file.",
```

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+            "description": "The size of the file in bytes.",
+            "type": "number",
+            "minimum": 0,
+            "user_friendly": "Size",
+            "example": 2147483648
+        },
+        "sha256": {
+            "description": "The sha256 hash of the file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{64}",
```
hannes-ucsc, line 70:
```suggestion
            "pattern": "^[a-f0-9]{64}",
```

Since so many people incorrectly perform a case-sensitive comparison, we shouldn't allow an ambiguous hexadecimal representation. Ambiguity is the mother of doom.

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+        "sha1": {
+            "description": "The sha1 hash of the file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{40}",
+            "user_friendly": "Sha1",
+            "example": "6e71b3cac15d32fe2d36c270887df9479c25c640"
+        },
+        "s3_etag" : {
+            "description": "The AWS S3 ETag for this file",
+            "user_friendly": "s3 ETag",
```
hannes-ucsc, line 90:
```suggestion
            "user_friendly": "S3 ETag",
```

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+        },
+        "sha1": {
+            "description": "The sha1 hash of the file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{40}",
+            "user_friendly": "Sha1",
+            "example": "6e71b3cac15d32fe2d36c270887df9479c25c640"
+        },
+        "s3_etag" : {
+            "description": "The AWS S3 ETag for this file",
```
hannes-ucsc, line 89:
```suggestion
            "description": "An AWS S3 ETag of this file",
```

Unfortunately, there could be many different ETags for any file, depending on whether it was multipart-uploaded or not and if it was, which part size was used.

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+            "description": "The CRC-32C code generated for this file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{8}",
+            "user_friendly": "Crc32c",
+            "example": "3FBA199E"
+        },
+        "sha1": {
+            "description": "The sha1 hash of the file.",
+            "type": "string",
+            "pattern": "^[A-Fa-f0-9]{40}",
```
hannes-ucsc, line 84:
```suggestion
            "pattern": "^[a-f0-9]{40}",
```

docs/jsonBrowser/core.md
```diff
@@ -18,6 +18,12 @@ file_name | The name of the file. | string | yes |  | File name |  | R1.fastq.gz
 format | The format of the file. | string | yes |  | File format |  | fastq.gz; tif
 content_description | General description of the contents of the file. | array | no | [See module  file_content_ontology](module.md/#file_content_ontology) | Content description |  | 
 checksum | MD5 checksum of the file. | string | no |  | Checksum |  | e09a986c2e630130b1849d4bf9a94c06
+content_type | An appropriate MIME type for this file. | string | no |  | Content type |  | application/gzip
```
hannes-ucsc, line 21: This markup is generated from the schema, right?

json_schema/core/file/file_core.json
```diff
@@ -48,7 +48,47 @@
...
+            "user_friendly": "Sha1",
```
hannes-ucsc, line 85:
```suggestion
            "user_friendly": "SHA-1",
```

docs/jsonBrowser/core.md
ESapenaVentura: Yep, Markups are generated automatically on every commit

ESapenaVentura:
>This deviates from the spec in two areas. @rolando-ebi had requested to nest the checksums in a file_provenance section and I worked that into the spec. This PR adds them to the root. I am in support of that deviation, but it should be proposed in the spec as well.

@hannes-ucsc I will update the specs to reflect this decision

> The second deviation is that this does not rename checksum to md5 which the spec requires. This deviation I am not exactly a fan of. I think consistency trumps backwards-compatibility in this case, especially considering that the property isn't actually used in the wild, or very infrequently used, according to @rolando-ebi. I am happy to discuss this further, but it would be prudent to have that discussion on a suggested change in the spec document.

I think the only reason to keep that field named as `checksum` would be to not break backwards compatibilty but, if that's not a problem, I am in favour of changing this field's name to md5. Happy to hear other's thoughts!

rolando-ebi:
@hannes-ucsc 
> This deviates from the spec in two areas. @rolando-ebi had requested to nest the checksums in a file_provenance section and I worked that into the spec. This PR adds them to the root. I am in support of that deviation, but it should be proposed in the spec as well.

The metadata team preferred to "unroll" these fields up into file-core rather than keep them use the "file_provenance" sub object. I think the unrolled version is what you originally proposed. I'm happy to do either, but yes we should reflect this in the spec as suggestions.

> checksum...md5

I'm also happy with this change

Suggested commits all look good to me

hannes-ucsc:
I agree that there is a trade-off to be made here. Just a little surprised where some of the comments appear to place the dial on the scale between compatibility and consistency.

> This sort of mass schema migration has never been done before.

Aren't we doing exactly that right now? For the DCP/2 MVP effort, the Azul team has already implemented four schema changes on hundreds of thousands of JSON documents as part of the DSS adapter. That would qualify as a mass migration in my book.

Given that our changes to `links.json` already broke backwards compatibility with v1, I would like to understand better what the specific issues are with renaming a rarely used and largely ignored field.

ESapenaVentura:
Since we have now another option for the file integrity checks in a different PR, should we close this @rolando-ebi @lauraclarke @hannes-ucsc @mshadbolt ?

mshadbolt:
yep

PR #1302, opened 2020-06-15
added file_descriptor.json schema

rolando-ebi:
This supports an [alternative proposal](https://docs.google.com/document/d/1NsibP8g-NeLnksxlcBWQsSj5Zg_uimCDaAF1qB_qkjg/edit?pli=1#heading=h.eeyjdjt197l9) for transmitting file checksums/integrity information

### DIFF
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -44,7 +44,7 @@
       "type": "string",
       "user_friendly": "File version",
       "example": "2019-09-24T093509.407370Z",
-      "format": "date-time"
+      "pattern": "\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}Z"
     },
     "content_type": {
       "description": "An appropriate MIME type for this file.",
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -3,6 +3,8 @@
   "description": "Checksums and other information for asserting the integrity of a data file.",
   "additionalProperties": false,
   "required": [
+    "describedBy",
+    "schema_type",
     "content_type",
     "size",
     "sha256",
@@ -25,6 +27,13 @@
       "type": "string",
       "pattern": "^[0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,}$",
       "example": "4.6.1"
+    },
+    "schema_type": {
+      "description": "The type of the metadata schema entity.",
+      "type": "string",
+      "enum": [
+        "file_descriptor"
+      ]
     },
     "schema_major_version": {
       "description": "The major version number of the schema.",
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -67,7 +67,7 @@
       "type": "string",
       "user_friendly": "File version",
       "example": "2020-05-01T04:26:07.021870Z",
-      "pattern": "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{6}Z"
+      "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{6}Z$"
     },
     "content_type": {
       "description": "An appropriate MIME type for this file.",
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -18,13 +18,27 @@
     "describedBy": {
       "description": "The URL reference to the schema.",
       "type": "string",
-      "pattern" : "^(http|https)://schema.(.*?)humancellatlas.org/system/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor"
+      "pattern" : "^(http|https)://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor"
     },
     "schema_version": {
       "description": "The version number of the schema in major.minor.patch format.",
       "type": "string",
-      "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+      "pattern": "^[0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,}$",
       "example": "4.6.1"
+    },
+    "schema_major_version": {
+      "description": "The major version number of the schema.",
+      "type": "integer",
+      "pattern": "^[0-9]{1,}$",
+      "user_friendly": "Schema major version",
+      "example": "4; 10"
+    },
+    "schema_minor_version": {
+      "description": "The minor version number of the schema.",
+      "type": "integer",
+      "pattern": "^[0-9]{1,}$",
+      "user_friendly": "Schema minor version",
+      "example": "6; 15"
     },
     "file_name": {
       "description": "The object name of the data file relative to the staging area's `data/` directory",
@@ -44,7 +58,7 @@
       "type": "string",
       "user_friendly": "File version",
       "example": "2020-05-01T04:26:07.021870Z",
-      "pattern": "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{6}Z"
+      "pattern": "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{6}Z"
     },
     "content_type": {
       "description": "An appropriate MIME type for this file.",
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -27,7 +27,7 @@
       "example": "4.6.1"
     },
     "file_name": {
-      "description": "The object name of the data file relative to the staging area",
+      "description": "The object name of the data file relative to the staging area's `data/` directory",
       "type": "string",
       "user_friendly": "Staging file name",
       "example": "1b6d8348-d6e9-406a-aa6a-7ee886e52bf9/IDC9_L004_R2.fastq.gz"
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -20,7 +20,7 @@
     "describedBy": {
       "description": "The URL reference to the schema.",
       "type": "string",
-      "pattern" : "^(http|https)://schema(\\.[a-z]+)?\\.humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
+      "pattern" : "^https?://schema(\\.[a-z]+)?\\.humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
     },
     "schema_version": {
       "description": "The version number of the schema in major.minor.patch format.",
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -20,7 +20,7 @@
     "describedBy": {
       "description": "The URL reference to the schema.",
       "type": "string",
-      "pattern" : "^(http|https)://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
+      "pattern" : "^(http|https)://schema(\\.[a-z]+)?\\.humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
     },
     "schema_version": {
       "description": "The version number of the schema in major.minor.patch format.",
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -6,7 +6,10 @@
     "content_type",
     "size",
     "sha256",
-    "crc32c"
+    "crc32c",
+    "file_id",
+    "file_version",
+    "file_name"
   ],
   "title": "File descriptor",
   "name": "file_descriptor",
@@ -24,7 +27,7 @@
       "example": "4.6.1"
     },
     "file_name": {
-      "description": "the object name of the data file relative to the staging area",
+      "description": "The object name of the data file relative to the staging area",
       "type": "string",
       "user_friendly": "Staging file name",
       "example": "1b6d8348-d6e9-406a-aa6a-7ee886e52bf9/IDC9_L004_R2.fastq.gz"
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -44,7 +44,7 @@
       "type": "string",
       "user_friendly": "File version",
       "example": "2020-05-01T04:26:07.021870Z",
-      "pattern": "\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}Z"
+      "pattern": "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{6}Z"
     },
     "content_type": {
       "description": "An appropriate MIME type for this file.",
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -20,7 +20,7 @@
     "describedBy": {
       "description": "The URL reference to the schema.",
       "type": "string",
-      "pattern" : "^(http|https)://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor"
+      "pattern" : "^(http|https)://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
     },
     "schema_version": {
       "description": "The version number of the schema in major.minor.patch format.",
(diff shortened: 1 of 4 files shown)


json_schema/system/file_integrity.json
```diff
@@ -0,0 +1,66 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "Checksums and other information for asserting the integrity of a data file.",
```
ESapenaVentura, line 3: Maybe something like `Checksums and data file integrity stats.`? "Other information" sounds vague to me

json_schema/system/file_integrity.json
```diff
@@ -0,0 +1,66 @@
...
+      "type": "string",
+      "pattern": "^[a-f0-9]{40}",
+      "user_friendly": "SHA-1",
+      "example": "6e71b3cac15d32fe2d36c270887df9479c25c640"
+    },
+    "s3_etag" : {
+      "description": "An AWS S3 ETag of this file",
+      "user_friendly": "S3 ETag",
+      "type": "string"
+    }
```
ESapenaVentura, line 64: Isn't it missing `md5`? or are we getting rid of it?

json_schema/system/file_integrity.json
```diff
@@ -0,0 +1,66 @@
...
+    "crc32"
+  ],
+  "title": "File integrity",
+  "name": "file_integrity",
+  "type": "object",
+  "properties": {
+    "describedBy": {
+      "description": "The URL reference to the schema.",
+      "type": "string",
+      "pattern" : "^(http|https)://schema.(.*?)humancellatlas.org/core/file/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/file_integrity"
```
hannes-ucsc, line 18:
```suggestion
      "pattern" : "^(http|https)://schema.(.*?)humancellatlas.org/system/file_integrity/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/file_integrity"
```

hannes-ucsc: Or, I don't know. But it seems core/file is wrong.

rolando-ebi, line 18: @hannes-ucsc good catch, this should probably be .../system/<version regex>/file_integrity. @ESapenaVentura can you confirm?

json_schema/system/file_integrity.json
rolando-ebi: The issue of renaming/moving checksum is orthogonal to this change in my mind. We could move md5 here but can't do so without considering the effect of moving/deleting checksum from file_core

json_schema/system/file_integrity.json
```diff
@@ -0,0 +1,66 @@
...
+      "pattern": "^[a-f0-9]{64}",
+      "user_friendly": "SHA-256",
+      "example": "12998c017066eb0d2a70b94e6ed3192985855ce390f321bbdb832022888bd251"
+    },
+    "crc32c": {
+      "description": "The CRC-32C code generated for this file.",
+      "type": "string",
+      "pattern": "^[a-f0-9]{8}",
+      "user_friendly": "CRC32C",
+      "example": "3FBA199E"
```
mshadbolt, line 51: @rolando-ebi the travis build is failing because the example given doesn't match the regex pattern, I tried looking for what a valid example or regex might be but couldn't find anything definitive, are you able to make these consistent?

rolando-ebi: @mshadbolt Ahh, just need to use lower case letters instead of uppercase. We previously allowed upper and lower case letters but agreed it's better to limit to to just lower case, as is convention

mshadbolt:
Noting that the spec now calls this json `file_descriptor.json` I will make this filename change just before merging in case it breaks any changes being made currently

json_schema/system/file_integrity.json
lauraclarke: Does leaving the checksum field in file.core cause any problems?

rolando-ebi: It shouldn't. It might be a source of confusion for data consumers who can see both file_core and the file_descriptor, but nothing will break by leaving it there

mshadbolt:
I added the additional fields that were requested in the DCP MVP schema doc. i.e. file_name, file_id, file_version

json_schema/system/file_integrity.json
```diff
@@ -60,7 +81,9 @@
     "s3_etag" : {
       "description": "An AWS S3 ETag of this file",
       "user_friendly": "S3 ETag",
-      "type": "string"
+      "type": "string",
+      "example": "c92e5374ac0a53b228d4c1511c2d2842-63",
+      "pattern": "^[a-f0-9]{32}-[a-f0-9]{2}"
```
rolando-ebi, line 86: @mshadbolt unfortunately, S3 Etags can have a varied structure depending on how a file was uploaded. From https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html: 

```
The entity tag is a hash of the object. The ETag reflects changes only to the contents of an object, not its metadata. The ETag may or may not be an MD5 digest of the object data. Whether or not it is depends on how the object was created and how it is encrypted as described below:

Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-S3 or plaintext, have ETags that are an MD5 digest of their object data.
Objects created by the PUT Object, POST Object, or Copy operation, or through the AWS Management Console, and are encrypted by SSE-C or SSE-KMS, have ETags that are not an MD5 digest of their object data.
If an object is created by either the Multipart Upload or Part Copy operation, the ETag is not an MD5 digest, regardless of the method of encryption.
Type: String
```

Rest of the fields look good.

mshadbolt: oh ok, I will remove the pattern.

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,90 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "description": "Checksums and other information for asserting the integrity of a data file.",
+  "additionalProperties": false,
+  "required": [
```
rolando-ebi, line 5: Do we also want fields like schema_type and describedBy to be required?

hannes-ucsc: Yes.

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,90 @@
...
+    "content_type",
+    "size",
```
hannes-ucsc, line 7: `size` would be redundant with `file_core` I'm leaning towards leaving it out.

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,90 @@
...
+      "pattern" : "^(http|https)://schema.(.*?)humancellatlas.org/system/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor"
+    },
+    "schema_version": {
+      "description": "The version number of the schema in major.minor.patch format.",
+      "type": "string",
+      "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+      "example": "4.6.1"
+    },
+    "file_name": {
+      "description": "The object name of the data file relative to the staging area",
```
hannes-ucsc, line 30:
```suggestion
      "description": "The object name of the data file relative to the staging area's `data/` directory",
```

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,90 @@
...
+      "user_friendly": "File UUID",
+      "example": "111f222a-333b-444c-555d-eee63ddfccc7",
+      "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+    },
+    "file_version": {
+      "description": "The version of the file given in date time format",
+      "type": "string",
+      "user_friendly": "File version",
+      "example": "2019-09-24T093509.407370Z",
+      "format": "date-time"
```
hannes-ucsc, line 47: I'd prefer to be very strict here and use the regex from the spec:

`\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6}Z`

json_schema/system/file_descriptor.json
mshadbolt: I don't understand this comment, there is no 'size' field in file_core

json_schema/system/file_descriptor.json
mshadbolt: isn't 'schema_type'  only relevant to type schemas? i.e. schemas that live in the 'type' directory. I don't know what the type of this schema would be. `system` ?

json_schema/system/file_descriptor.json
mshadbolt: pushed a commit that removed format and added regex pattern as above

mshadbolt: after pushing realised the regex provided doesn't match the example for file_version in the spec. @rolando-ebi are you able to confirm the format of the version and provide an accurate regex?

json_schema/system/file_descriptor.json
hannes-ucsc: I got confused. You are right, `size` needs to be in the `file_descriptor` schema.

json_schema/system/file_descriptor.json
hannes-ucsc: It would be `file_descriptor`, just as for `links.json` it is `links`.

https://github.com/HumanCellAtlas/metadata-schema/blob/cf8a4898aa4d9663eb6f19357bff22ecbb2dec4b/json_schema/system/links.json#L233

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,90 @@
...
+      "type": "string",
+      "user_friendly": "File UUID",
+      "example": "111f222a-333b-444c-555d-eee63ddfccc7",
+      "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+    },
+    "file_version": {
+      "description": "The version of the file given in date time format",
+      "type": "string",
+      "user_friendly": "File version",
+      "example": "2019-09-24T093509.407370Z",
```
hannes-ucsc, line 46:
```suggestion
      "example": "2020-05-01T04:26:07.021870Z",
```

json_schema/system/file_descriptor.json
mshadbolt: ok cool. resolving this one

json_schema/system/file_descriptor.json
mshadbolt: ok we can add it but from my understanding it isn't normally how we use this field. as an example with `donor_organism`, the donor_organism **type** schema has the `schema_type` field as "biomaterial" whereas the core and module schemas don't have this field, e.g. `biomaterial_core`, `medical_history`. I guess I don't see the purpose of having a 'type' when there is only one schema with that type? Not sure I understand why it is a part of the `links.json` either. anyways if it is useful for something then we can add it.

hannes-ucsc: I'm not sure either what the intended use of that field is either, but to reduce risk for *this* PR, it would make sense to be consistent with the other system schemas. If we plan on eliminating `schema_type` from the `system` schemas, I would do that to all of them at the same time, in a different PR.

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,113 @@
...
+      "user_friendly": "File UUID",
+      "example": "111f222a-333b-444c-555d-eee63ddfccc7",
+      "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+    },
+    "file_version": {
+      "description": "The version of the file given in date time format",
+      "type": "string",
+      "user_friendly": "File version",
+      "example": "2020-05-01T04:26:07.021870Z",
+      "pattern": "[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{6}Z"
```
hannes-ucsc, line 70:
```suggestion
      "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\\.[0-9]{6}Z$"
```

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,113 @@
...
+      "description": "The size of the file in bytes.",
+      "type": "number",
+      "minimum": 0,
+      "user_friendly": "Size",
+      "example": 2147483648
+    },
+    "sha256": {
+      "description": "The SHA-256 hash of the file.",
+      "type": "string",
+      "pattern": "^[a-f0-9]{64}",
```
hannes-ucsc, line 88:
```suggestion
      "pattern": "^[a-f0-9]{64}$",
```

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,113 @@
...
+    "file_name"
+  ],
+  "title": "File descriptor",
+  "name": "file_descriptor",
+  "type": "object",
+  "properties": {
+    "describedBy": {
+      "description": "The URL reference to the schema.",
+      "type": "string",
+      "pattern" : "^(http|https)://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor"
```
hannes-ucsc, line 23:
```suggestion
      "pattern" : "^(http|https)://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
```

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,113 @@
...
+      "description": "The SHA-256 hash of the file.",
+      "type": "string",
+      "pattern": "^[a-f0-9]{64}",
+      "user_friendly": "SHA-256",
+      "example": "12998c017066eb0d2a70b94e6ed3192985855ce390f321bbdb832022888bd251"
+    },
+    "crc32c": {
+      "description": "The CRC-32C code generated for this file.",
+      "type": "string",
+      "pattern": "^[a-f0-9]{8}",
```
hannes-ucsc, line 95:
```suggestion
      "pattern": "^[a-f0-9]{8}$",
```

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,113 @@
...
+      "user_friendly": "CRC32C",
+      "example": "0b83b575"
+    },
+    "sha1": {
+      "description": "The SHA-1 hash of the file.",
+      "type": "string",
+      "pattern": "^[a-f0-9]{40}",
```
hannes-ucsc, line 102:
```suggestion
      "pattern": "^[a-f0-9]{40}$",
```

json_schema/system/file_descriptor.json
mshadbolt: good point, the more I look at our regexes the more issues I find :( thanks for the commit and we will do a general review as well.

mshadbolt:
I believe this PR is ready to approve if @hannes-ucsc can check overnight and @rolando-ebi and @ESapenaVentura thinks it looks ok I can hot fix tomorrow

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,113 @@
...
+    },
+    "content_type": {
+      "description": "An appropriate MIME type for this file.",
+      "type": "string",
+      "user_friendly": "Content type",
+      "example": "application/gzip"
+    },
+    "size": {
+      "description": "The size of the file in bytes.",
+      "type": "number",
```
mshadbolt, line 80: should "type" here actually be "integer" instead of "number"?

rolando-ebi: @mshadbolt Good catch, yep this should be integer

rolando-ebi, line 80:
```suggestion
      "type": "integer",
```

json_schema/system/file_descriptor.json
```diff
@@ -0,0 +1,113 @@
...
+    "file_name"
+  ],
+  "title": "File descriptor",
+  "name": "file_descriptor",
+  "type": "object",
+  "properties": {
+    "describedBy": {
+      "description": "The URL reference to the schema.",
+      "type": "string",
+      "pattern" : "^(http|https)://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
```
ESapenaVentura, line 23: If we are being super strict with the regex expressions, I think the pattern should be:

`^https?://schema\\.(dev\\.|integration\\.|staging\\.)?humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$`

This should be enough to not match anything that is not a correct URL

Although now that I look at it, what is the ...`|([a-zA-Z]*?)`almost at the end of the regex? I thought our versioning is exclusively done with numbers?

hannes-ucsc, line 23:
```suggestion
      "pattern" : "^(http|https)://schema(\\.[a-z]+)?\\.humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
```

- is a bit stricter
- `.*?` is the same as `.*`
- `.*` in general should be avoided in URL patterns since it can be very easily exploited to point to a completely different host. Think `http://schema.badhost.io/#.humancellatlas.org/system/1.2.3/file_descriptor`

hannes-ucsc, line 23:
```suggestion
      "pattern" : "^https?://schema\\.(.*?)humancellatlas\\.org/system/(([0-9]{1,}\\.[0-9]{1,}\\.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
```

Nit

mshadbolt: done

mshadbolt:
I am going to close this pull request because the base branch was from develop it will cause issues when I try to hotfix into master so I created a new branch (https://github.com/HumanCellAtlas/metadata-schema/tree/ms-file_descriptor-%231289) with the exact same json that I will use to hotfix.
See the new PR here: #1312

PR #1319, opened 2020-06-26
Remove major/minor version fields from file_descriptor & provenance. Fixes #1317

### LINKED ISSUE #1317: Remove `schema_major_version` and `schema_minor_version` from file_descriptor
We should not have added these to file_descriptor in the first place. We've already agreed to phase them out and they are absent from other system schemas like `links.json`.

From [DCP v2 spec, section Schema validation](https://docs.google.com/document/d/1NsibP8g-NeLnksxlcBWQsSj5Zg_uimCDaAF1qB_qkjg/edit#heading=h.oatkw4xqqy0o):

> There is one known existing schema violation in documents in the DCP v1 production instance of DSS. The `provenance.schema_major_version` and `provenance.schema_minor_version` properties are present in metadata files submitted to the DSS from around Oct 2019 onwards. The addition of these fields was proposed in [RFC 11](https://github.com/HumanCellAtlas/dcp-community/blob/master/rfcs/text/0011-query-by-metadata-schema-versions.md). After that RFC was accepted, the `provenance` schema was revised and the Ingest component was modified to add those fields to documents but the schema reference in those documents still points at an old schema, not the revised one. This is further complicated by the fact that the `provenance` schema is referenced indirectly via the main document schema ([example document](http://dss.data.humancellatlas.org/v1/files/003d3dda-6906-4943-9f12-331b963e2f55?replica=aws)). The `analysis_file` schema in that document is at version 6.0.0 while the new fields were introduced in version 6.2.0 (via the `provenance` schema version 1.1.0). The problem affects all metadata documents submitted after October of 2019, not just `analysis_file` documents.

> To address this issue, the DSS adapter removes those two fields and the Ingest adapter is modified to not emit them. Luckily, the fields are not required so removing them from documents that do happen to carry an updated schema declaration does not invalidate those documents. The `provenance` schema will be revised to remove the fields again.

These two properties need to be removed from the schema:

https://github.com/HumanCellAtlas/metadata-schema/blob/c2ab8a5115a2fabd4fcbb57fda89022be12cd939/json_schema/system/file_descriptor.json#L38

https://github.com/HumanCellAtlas/metadata-schema/blob/c2ab8a5115a2fabd4fcbb57fda89022be12cd939/json_schema/system/file_descriptor.json#L45

mshadbolt:
### Release notes

For file_descriptor schema:
- Removed `schema_major_version`
- Removed `schema_minor_version`

### Reviews requested

- Last Reviewer to review/approve, merge changes by Thursday 2nd July.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,10 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
 
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/develop)
 
+### [system/file_descriptor.json - v2.0.0] - 2020-06-29
+### Removed
+Removed schema major minor versions.
+
 ## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
 
 ### [system/file_descriptor.json - v1.0.0] - 2020-06-25
--- a/changelog.md
+++ b/changelog.md
@@ -7,11 +7,11 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
 
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/develop)
 
+## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
+
 ### [system/file_descriptor.json - v2.0.0] - 2020-06-29
 ### Removed
 Removed schema major minor versions.
-
-## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
 
 ### [system/file_descriptor.json - v1.0.0] - 2020-06-25
 ### Added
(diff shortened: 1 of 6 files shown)


mshadbolt:
@ESapenaVentura are you particularly able to review the property_migrations file? (on monday) I am not sure which fields I am/am not supposed to fill based on the documentation.

mshadbolt:
yep both major version changes

mshadbolt:
I had to revert the changes to the provenance schema because it affects every type schema. 

We will still ensure our exporter does not export metadata files with the provenance major/minor versions to DCP2 but we will do this change at a later date once we are sure it doesn't affect the validity of metadata in ingest.

PR #1326, opened 2020-07-01
Revert INSDC project accession change but keep newer commits

mshadbolt:
This isn't an official release. We decided to revert the INSDC project/study accession change because it greatly complicates our release process. In this PR I reverted this change while incorporating newer commits that are either already released on master or are safe to be promoted to the other branches, i.e. documentation updates or snyk PRs.


ESapenaVentura:
@mshadbolt please check the info in here https://github.com/HumanCellAtlas/metadata-schema/blob/ms-develop-revert/docs/jsonBrowser/type.md is correct. 

You can run `src/human_readable_json.py` to make sure :)

ESapenaVentura:
LGTM, cannot think of anything else. I have also tried to compare this branch with master and it looks like the only changes are not related to the schema (docs and snyk), so I think it's good to go

PR #1359, opened 2021-02-26
Add identifier, url modules and fields to type/project, Fixes #1336

### LINKED ISSUE #1336: Enable capture of portal links and SCEA, EGA & dbGap accessions

project.json

I would like to add at least two optional fields to the project_core.json schema

#### ega accession
* Field name: ega_accessions
* Field description: European Genome Phenome Archive (EGA) dataset or study accession. Study accession preferred.
* Field type: list of strings
* Required: No
* Examples: EGAS00000000001; EGAD00000000002
* pattern: `EGA[DS][0-9]{11}`

#### dbgap accession
* Field name: dbgap_accessions
* Field description: database of Genotypes and Phenotypes (dbGaP) study accession.
* Field type: list of strings
* Required: No
* Examples: phs001997.v1.p1; phs001836
* pattern: `phs[0-9]{6}(\.v[0-9]{1})?(\.p[0-9])?` (this regex should be double checked)

#### scea accession
* Field name: scea_accessions
* Field description: Single cell expression atlas accession.
* Field type: list of strings
* Required: No
* Examples: E-MTAB-6678;E-GEOD-150728
* pattern: `E-[A-Z]{4}-[0-9]+` (this regex should be double checked)

#### contributor portal links
* Field name: portal_links
* Field description: Links to a contributor created web-based portal with visualisations for this project.
* Field type: list of strings
* type: string
* format: uri
* Required: No
* Examples: https://www.heartcellatlas.org/
* pattern: `^https?://.+` (this regex should be double checked)

To support additional links to managed access archives when that is where the data is available and contributor portal links.

mshadbolt:
I am making this PR as a starting point to get feedback. I am not sure if we want to remove all the existing accession and links fields yet or if we want to have a soft transition while we figure out a plan of migration. 

### Release notes

Created External Identifier module as a non typed module

Created URL module as a project module

For type/project schema:
- added optional project_accessions field
- added optional project_urls field

### Reviews requested

- Last Reviewer to review/approve, merge changes by Friday, 5th March

### DIFF
--- a/docs/jsonBrowser/module.md
+++ b/docs/jsonBrowser/module.md
@@ -7,6 +7,17 @@ Property name | Description | Type | Required? | Example
  describedBy | The URL reference to the schema. | string | no |  |  |  | 
 schema_version | The version number of the schema in major.minor.patch format. | string | no | 4.6.1
 
+## External Identifier<a name='External Identifier'></a>
+_Information about an external identifier._
+
+Location: module/external_identifier.json
+
+Property name | Description | Type | Required? | Object reference? | User friendly name | Allowed values | Example 
+--- | --- | --- | --- | --- | --- | --- | --- 
+external_identifier | Any unique identifier that identifies the project, dataset, experiment, biomaterial or protocol. | string | yes |  | External Identifier |  | SCP464; E-MTAB-8901
+identifier_source | The source of the unique identifier. | string | yes |  | Identifer Source | insdc_study, insdc_project, insdc_sample, insdc_experiment, biosamples, indsc_run, ega_study, ega_dataset, ega_sample, arrayexpress, scea, scp, geo_series, dbgap, cbtm, hdbr, biostudies, other | 
+other_explanation | An explanation of the source of the identifier. | string | no |  | Other explanation |  | Identifier used for internal tracking.
+
 ## Channel<a name='Channel'></a>
 _Information about a single microscope channel._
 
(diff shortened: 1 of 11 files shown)


mshadbolt:
schema_linter is failing because it doesn't like my use of the 'anyOf' property in the 'external_identifier.json' to attempt to validate each value of the enum. 

@ESapenaVentura are you able to do a sense check to see if you think I implemented it properly based on the example you shared with me?

I guess then we will need to check if there is any good reason not to include the anyOf property as an allowed property and whether our json schema validator will successfully validate these.

ESapenaVentura:
Hi @mshadbolt ! I have checked the external_identifier.json

The field "allOf" needs to be outside properties! Aside from that, I needed to make a small change to the accepted schema keywords in the schema linter

I have checked the external_identifiers.json schema here https://www.jsonschemavalidator.net/ agains this
```
{
	"external_identifier": "E-MTAB-8901",
    "identifier_source" : "arraexpress"
}
```
And it works!

I still need to check other aspects of the schema changes though

By the way, it looks like you don't have pre-commit installed, as only when I pushed these commits it generated the markdown files for these new schemas

ESapenaVentura:
Also I wonder, do we need to create a "general" module folder? I understand that this accession module can be applied to every type entity and you didn't put it there for that reason, right?

mshadbolt:
ok thanks @ESapenaVentura !

I am not sure the best location for it as yes I think it would be good if this module can be used by any type schema. I chose to put it in the root module folder but if you think it would be less confusing to have it in its own folder we could do that too.

Not 100% sure on how we will use this module to replace all the existing identifier fields we have  at the moment. e.g. it could replace the 'hdbr' id and allow a user to specify a cbtm id. I was a bit undecided on whether to add all the potential options for the types of identifiers at this stage or only have the project level ones for now so as not to be confusing.

I might try to add some tests to ensure that these enums are validating the way that we expect.

mshadbolt:
I added a few tests though I think we need to establish a testing strategy. For every change that we make do we want to have a pass/fail test?

It also occurred to me that to have proper fail tests we would need to have a separate fail for each thing we want to fail, because if there are multiple fails in one fail test it only checks the first one and fails it. I would need to check the code to see whether it just automatically checks every file in the tests folder or any additional configuration needs to be done.

lauraclarke:
Making it work for project accessions before rolling out to any other entity type makes sense, unless there is any advantage to being able to store CBTM identifiers as well as HDBR ones in the immediate future?

mshadbolt:
I think this change is too complex to support right now so will close this PR in favour of making minor schema changes to the support at the project level for now

PR #1363, opened 2021-03-04
Added test metadata file

ESapenaVentura:
Added the test metadata spreadsheet

Not tested yet.

Should contain everything, if @mshadbolt or any other wrangler wants to take a look it would be great

Tomorrow I'll fill this PR better, this is in case @mshadbolt wakes up and wants to test/check/add


ESapenaVentura:
Note: I have not generated the files yet

ESapenaVentura:
I have generated the files, filled in the spreadsheet and tested in ingest, it works fine.

Imaging tabs needed to be taken out of this first iteration so there are 2 spreadsheets to make sure when we are able to ingest the imaging tabs we can modify the spreadsheet faster

@mshadbolt has reviewed it

I am going to merge and include in the integration release

Issue #1333, opened 2020-10-23
`protocol_type` in links schema should be an enum


hannes-ucsc:
… with an entry for each of the valid schema types for protocols.

https://github.com/HumanCellAtlas/metadata-schema/blob/90b0ac22def50d35538891c6d145d07077a8f93a/json_schema/system/links.json#L71

Would have prevented https://humancellatlas.slack.com/archives/C01360XN04S/p1603433805210700

ESapenaVentura:
Closing as this has been merged to develop

PR #1371, opened 2021-03-17
Release from staging to master - 2021-03-17

ESapenaVentura:
### Release notes

**Versions:**

system/links.json - v3.0.0
module/protocol/matrix.json - v1.0.0
type/protocol/analysis/analysis_protocol.json - v9.2.0
type/file/analysis_file.json - v6.3.0
core/file/file_core.json - v6.2.0
type/file/sequence_file.json - v9.3.0
type/file/supplementary_file.json - v2.3.0
type/file/reference_file.json - v3.3.0
type/file/image_file.json - v2.3.0
module/ontology/data_use_ontology.json - v1.0.0
type/project/project.json - v15.0.0

**Test metadata**:
https://github.com/HumanCellAtlas/schema-test-data/pull/8

**Functionality:**

- Fixed javascript validator (Points to correct URL)
- Added fail_test for links.json
- Fixed schema linter (Points to correct URL)
- Fixed pullapprove and updated lists
- Added test data spreadsheet

**Additional notes:**
- `type/project/project.json` has advanced 1 major version because of a rollback. The major change should have no impact as it involves the deletion of a field that was never used.


docs/jsonBrowser/module.md
```diff
@@ -39,6 +39,16 @@ assay_type | Type of assay used to detect target. | object | yes | [See module
...
 channel_label | Channel label used to assay signal. | array | no |  | Channel |  | 1; A
 
+## Matrix<a name='Matrix'></a>
+_Information relating to generation of count matrices_
+
+Location: module/protocol/matrix.json
+
+Property name | Description | Type | Required? | Object reference? | User friendly name | Allowed values | Example 
+--- | --- | --- | --- | --- | --- | --- | --- 
+data_normalization_methods | Method(s) used to normalize data in the matrix | array | no |  | Data normalization method(s) |  |
```
ruchim, line 49: I'd love to see a list of allowed "type"s in the docs. It will help our validation, plus it will make it clear to the consumers of this standard for what's appropriate!

docs/jsonBrowser/module.md
```diff
@@ -83,6 +93,17 @@ text | The name of a library amplification approach being used. | string | yes |
 ontology | An ontology term identifier in the form prefix:accession. | string | no |  | Library amplification ontology ID |  | OBI:0000415; EFO:0009013
 ontology_label | The preferred label for the ontology term referred to in the ontology field. This may differ from the user-supplied value in the text field. | string | no |  | Library amplification ontology label |  | PCR; in vitro transcription
 
+## Data use ontology<a name='Data use ontology'></a>
```
ruchim, line 96: hmm so I think earlier we said we'd remove it, so this feels counter to that decision?

json_schema/core/file/file_core.json
```diff
@@ -49,6 +49,17 @@
             "type": "string",
             "user_friendly": "Checksum",
             "example": "e09a986c2e630130b1849d4bf9a94c06"
-	    }
+	    },
+        "file_source": {
+            "description": "Has the file come from a contributor or the DCP?",
```
ruchim, line 54: (optional) Alternat description: The name of the entity that generated the file

json_schema/module/ontology/data_use_ontology.json
```diff
@@ -0,0 +1,49 @@
+{
```
ruchim, line 1: I don't believe this file should be included!

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
...
+                "UQ (Upper quartile)",
+                "Downsampling",
+                "other",
+                "unknown"
+            ]
+            },
+            "guidelines": "Should be one of: CPM (counts per million), TPM (transcripts per kilobase million), RPKM(reads per kilobase of exon per million reads mapped), FPKM (fragments per kilobase of exon per million fragments mapped), DESeq2’s median of ratios, TMM (EdgeR’s trimmed mean of M values), SF (size factor), UQ (Upper quartile), Downsampling, Unknown, other"
+        },
+        "derivation_process": {
+            "description": "How is the data derived?",
```
ruchim, line 43: given the items, it seems the derivation process is more like "type of computational tool used in generating the matrix object"

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
...
+            "type": "array",
+            "user_friendly": "Derivation process",
+            "guidelines": "Should be one of: alignment, quantification, peak calling, peak annotation, gene filtering, cell filtering, merging, cell calling, ambient RNA correction, doublet removal, batch correction, depth normalization",
+            "items": {
+                "type": "string",
+                "enum": [
```
ruchim, line 49: question:so is this field a list of enums or a single enum. Meaning can items = ["alignment", "quantification"] or it has to be items = "alignment"

json_schema/property_migrations.json
```diff
@@ -1,5 +1,12 @@
 {
     "migrations": [
+        {
+            "source_schema": "project",
+            "property": "data_use_restrictions",
```
ruchim, line 5: I'd love a description of what this enables? data use restrictions feels a bit like arbitrary terminology and if we're removing DUO codes, not sure this is necessary

json_schema/type/project/project.json
```diff
@@ -77,6 +77,28 @@
             "user_friendly": "INSDC project accession",
             "guidelines": "Enter accession if project has been archived. Accession can be from the DDBJ, NCBI, or EMBL-EBI and must start with DRP, SRP, or ERP, respectively."
         },
+        "ega_accessions": {
+            "description": "A list of European Genome phenome Archive dataset or study accessions.",
```
ruchim, line 81: this is a list because 1 HCA project = many EGA studies?

ruchim: sorry, you probably mentioned this but just verifying to be absolutely sure

tests/schema_test_files/system/test_fail_links.json
```diff
@@ -0,0 +1,47 @@
+{
```
ruchim, line 1: given the name of the file -- is this a test data to show a bad links json? can we add a comment to describe what causes it to fail?

ruchim (changes requested):
a) Requesting changes/clarification around DUO changes. b) A few questions about about why certain changes were made  c) a few optional thoughts about reframing descriptions

Looking forward to re-reviewing!

docs/jsonBrowser/module.md
ESapenaVentura: We decided to keep the module but there is no field in the schema that loads it, so it's effectively inexistent! (Can't be used in a project)

We took this decision to avoid some work in the future re-creating the module, but we can discuss if we need to delete it completely

json_schema/module/protocol/matrix.json
ESapenaVentura: it's a list of enums, meaning that e.g. "alignment" and "quantification" can be both specified for the same protocol

json_schema/type/project/project.json
ESapenaVentura: Yep! there may be cases where 1 HCA project == Many EGA accessions

docs/jsonBrowser/module.md
ruchim: if you're worried about recreating the code, maybe it can just be put in a separate branch? @hannes-ucsc does that feel right to you too?

json_schema/property_migrations.json
ESapenaVentura: This is necessary for an automated migration tool, this will be used (Still not operational) to try and automate migrations to the best of our extents.

This just states that from project 14 (last major update, recorded down in the document) to 15, the field was deleted. (Basically, if someone was to update from 14.3.0 to 15.0.0, they need to delete the field to make the instance valid)

Of course, this has been a rollback so technically it has never been used and no project entity will need this specific update, but I think it's best practice to keep a record of this anyways!

docs/jsonBrowser/module.md
ESapenaVentura: I agree, this is usually the case but thanks to this comment I have observed there's a bug where enum arrays won't be loaded properly. I'm opening a ticket to investigate into this issue!

ESapenaVentura: Fixed it in this PR https://github.com/HumanCellAtlas/metadata-schema/pull/1379/files since it's a webhook it actually also fixed this doc :)

json_schema/core/file/file_core.json
ESapenaVentura: I think that may be a little confusing to some users (e.g. people may understand that field as the institution they are working in), but happy to discuss about it

hannes-ucsc: The term entity is already taken.

The phrasing is a bit awkward but more importantly, I'd avoid repeating the enum values in the description. We'd have to rewrite it every time a new file source needs to be added.

```suggestion
            "description": "The source of the file. This is typically an organisation, person or dedicated process.",
```

json_schema/module/ontology/data_use_ontology.json
hannes-ucsc: Correct. As agreed on 3/24.

json_schema/module/protocol/matrix.json
hannes-ucsc: Is there not an existing ontology we can utilize? Part of the reason we're doing this whole thing is to rely on EBI's ontology validation.

json_schema/property_migrations.json
hannes-ucsc: No data use related changes should be in this PR. Not even rollback migrations of features already released. This needs to be removed.

json_schema/type/file/analysis_file.json
```diff
@@ -39,6 +39,15 @@
             "type": "object",
             "$ref": "core/file/file_core.json",
             "user_friendly": "File core"
+        },
+        "matrix_cell_count": {
+            "description": "Number of cells analyzed in the matrix.",
```
hannes-ucsc, line 44: Not all analysis files are matrices. The description should clarify that this property only applies to the latter.

json_schema/type/project/project.json
hannes-ucsc: Capitalization and phrasing is odd.
```suggestion
            "description": "A list of accessions referring to EGA (European Genome-Phenome Archive) datasets or studies.",
```

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
...
+            "example": "4.6.1"
+        },
+        "data_normalization_methods": {
+            "description": "Method(s) used to normalize data in the matrix",
+            "type": "array",
+            "user_friendly": "Data normalization method(s)",
+            "items":{
+                "type": "string",
+                "enum": [
+                "CPM (counts per million)",
```
hannes-ucsc, line 27:
```suggestion
                    "CPM (counts per million)",
```

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
...
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+        "data_normalization_methods": {
+            "description": "Method(s) used to normalize data in the matrix",
+            "type": "array",
+            "user_friendly": "Data normalization method(s)",
+            "items":{
+                "type": "string",
+                "enum": [
```
hannes-ucsc, line 26: Same question, is there no existing ontology we could use?

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
...
+                "DESeq2’s median of ratios",
+                "TMM (EdgeR’s trimmed mean of M values)",
+                "SF (size factor)",
+                "UQ (Upper quartile)",
+                "Downsampling",
+                "other",
+                "unknown"
+            ]
+            },
+            "guidelines": "Should be one of: CPM (counts per million), TPM (transcripts per kilobase million), RPKM(reads per kilobase of exon per million reads mapped), FPKM (fragments per kilobase of exon per million fragments mapped), DESeq2’s median of ratios, TMM (EdgeR’s trimmed mean of M values), SF (size factor), UQ (Upper quartile), Downsampling, Unknown, other"
```
hannes-ucsc, line 40: Did you write this by hand? If so why can't this be generated automatically? (I assume this is predominantly meant for people filling out spreadsheets),

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Information relating to generation of count matrices",
+    "additionalProperties": false,
+    "title": "Matrix",
+    "name": "matrix",
+    "type": "object",
+    "properties": {
```
hannes-ucsc, line 8: Which of these properties would you expect to be exposed in the Data Browser?

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "description": "Information relating to generation of count matrices",
```
hannes-ucsc, line 3: Sure this is only for count matrices? What about GEX matrices?

json_schema/system/links.json
```diff
@@ -72,7 +72,20 @@
                     "description": "The concrete type of protocol entity.",
                     "example": "enrichment_protocol",
                     "user_friendly": "Protocol type",
-                    "type": "string"
+                    "type": "string",
+                    "enum": [
```
hannes-ucsc, line 76: I'm assuming this is an unrelated bug fix?

json_schema/versions.json
```diff
@@ -1,12 +1,12 @@
 {
```
hannes-ucsc, line 1: What's the purpose of this file? Neither TDR nor Azul rely on it.

hannes-ucsc (changes requested):
The data use stuff needs to go.

Off-topic: 

Reviewing this PR is made more difficult by the cruft contained therein. The documentation markup is automatically generated so ideally it wouldn't be part of the history. I don't really understand why an artifact that can be generated *from* the source, needs to be part *of* the source. The history in the PR is incomprehensible. No reviewer is gonna go though 114 commits.

I'm fine with letting these cruft issues slide the first time but it is not sustainable in the long run. Any ideas of how this could be improved?

json_schema/versions.json
mshadbolt: It is used by the scripts within this repo to increment version numbers when we run the 'release_prepare.py' script. The following lines in the 'version_update.py' reads this json file to get current versions then writes the updated `versions.json` with the updated versions of any schemas or dependent schemas, that have been incremented by the major/minor/patch release. (https://github.com/HumanCellAtlas/metadata-schema/blob/ff27788edc635f6e631a64b3fcda63f54fc6fe7b/src/version_update.py#L25, https://github.com/HumanCellAtlas/metadata-schema/blob/ff27788edc635f6e631a64b3fcda63f54fc6fe7b/src/version_update.py#L64). 

I believe it is also used by our [metadata schema publisher](https://github.com/HumanCellAtlas/metadata-schema-publisher) that publishes the versioned schemas to their url addresses.

tests/schema_test_files/system/test_fail_links.json
mshadbolt: I am not sure I feel that comfortable adding comments to json files (though I realise this might be controversial). There are a set of basic tests including json validation and some pass/fail tests that are run by Travis each time a PR is made (or run manually as we make changes). In the output from these tests, one can see the reason why a fail test fails: https://travis-ci.com/github/HumanCellAtlas/metadata-schema/jobs/494698958#L400

json_schema/system/links.json
mshadbolt: This was a change you had requested. We thought since we had already merged it and it has zero impact it would be okay to include in this release. https://github.com/HumanCellAtlas/metadata-schema/issues/1333

docs/jsonBrowser/module.md
Wkt8: We will remove this from the PR.

json_schema/module/protocol/matrix.json
Wkt8: There is not an existing ontology at the moment. However, we think there is still value added from this field to data consumers, even if it is not ontologised.

json_schema/module/ontology/data_use_ontology.json
Wkt8: We will remove it from the PR.

json_schema/module/protocol/matrix.json
Wkt8: Same answer, there is not one now but we think this field still adds enough value without one.

json_schema/module/protocol/matrix.json
Wkt8: That sounds like a sensible change we can make to the description field. We will do it.

json_schema/type/file/analysis_file.json
Wkt8: That is correct. Will make this change.

json_schema/module/protocol/matrix.json
Wkt8: Good catch. We'll make this change and make it the more general 'matrices'.

mshadbolt:
Just so the reviewers who have reviewed so far are aware (@hannes-ucsc @ruchim). We have taken your feedback on board and we are now preparing a new feature branch to merge into staging with your suggested fixes. We are also removing all traces of the data_use_ontology_module and rolling back the project versioning and re-releasing the schemas so that it was like the module and the version of the project schema that referred to it never existed. We will also re-generate the test data. Once this is all completed we will re-request review. Thanks for all your feedback.

json_schema/module/protocol/matrix.json
mshadbolt: I made an assessment that the time it would take me to figure out how to generate this automatically would be longer than doing a quick manual hack to replace characters and reformat into a sentence. Thanks for catching the mistake that I made. Yes it is displayed in the spreadsheet so that users can see what options they are allowed to use for the field.

json_schema/type/protocol/analysis/analysis_protocol.json
```diff
@@ -53,6 +52,12 @@
             "type": "string",
             "user_friendly": "Computational method",
             "example": "SmartSeq2SingleCell; 10x"
+        },
```
kbergin, line 55: Is this a required field?

kbergin (commented):
Had a question about the analysis_protocol json changes. Outside of that question, everything else shouldn't impact analysis as far as I can tell.

json_schema/module/protocol/matrix.json
hannes-ucsc: I didn't catch any mistakes. I just noticed that it's a bad idea to manually maintain this field if it can be generated automatically from the enum values. It's not fair to compare the effort one manual fix to that of an automatic implementation, if that implementation eliminates the need to maintain the duplication for *every* enum in the schema, and *every* time an enum is modified. At some point the automatic implementation effort will amortize. I won't hold up the PR for this reason. It's a minor annoyance to the reviewers because it increases the amount of text to review.

json_schema/module/protocol/matrix.json
hannes-ucsc: Let's continue this discussion here: https://github.com/HumanCellAtlas/dcp2/pull/8#discussion_r604326828

json_schema/versions.json
hannes-ucsc: Got it. Thank you for the explanation.

json_schema/system/links.json
hannes-ucsc: I don't think this is zero impact. If there is a typo or omission in this enum, it could cause future exports to fail to import into TDR.

When you say "we had already merged it" do you mean merged into develop and staging? If so, that statement applies to all changes in this PR, correct?

Are there any staging areas in the wild containing `links.json` documents that use this schema?

I'm just trying to evaluate the risks that lumping this unrelated change in would create.

json_schema/type/protocol/analysis/analysis_protocol.json
hannes-ucsc: No. See the `required` property in line 5 of this file.

json_schema/type/file/analysis_file.json
```diff
@@ -39,6 +39,15 @@
             "type": "object",
             "$ref": "core/file/file_core.json",
             "user_friendly": "File core"
+        },
+        "matrix_cell_count": {
```
hannes-ucsc, line 43: @kbergin, do you plan to support this field in the `analysis_file` instances created by your workflows?

json_schema/type/protocol/analysis/analysis_protocol.json
hannes-ucsc: Do you plan to support any of the properties added by the matrix module in the `analysis_protocol` instances created by your workflows?

json_schema/core/file/file_core.json
```diff
@@ -49,6 +49,17 @@
...
             "example": "e09a986c2e630130b1849d4bf9a94c06"
-	    }
+	    },
+        "file_source": {
+            "description": "The source of the file. This is typically an organisation, person or dedicated process.",
+            "type": "string",
+            "enum":[
+                "DCP",
+                "Contributor"
+            ],
```
hannes-ucsc, line 59:
```suggestion
            "enum":[
                "DCP/2 Analysis",
                "Contributor",
                "Array Express",
                "GEO",
                "SCEA",
                "SCP",
                "DCP/1 Matrix Service",
                "LungMAP",
                "Zenodo",
                "Publication",
            ],
```

to be consistent with the set of sources that we currently track (https://github.com/DataBiosphere/azul/blob/6b45e0e30fbe90c7b3944ee691c95328c7883ab2/src/azul/plugins/metadata/hca/transform.py#L271).

It would be undesirable to be less informative than the current "hacky" mechanism.

hannes-ucsc (changes requested):
A few more comments.

json_schema/core/file/file_core.json
hannes-ucsc: Might also want to add `"HCA Release"`

json_schema/system/links.json
mshadbolt: I think one of the aims of having all components review a change to the schema is to catch any typos or omissions. So while we have put our best effort into capturing all the values for these fields accurately in the enum, and reviewed it internally, I would hope that the components that rely on these fields would also check to see that we haven't omitted or misspelled anything.

If you prefer us to remove this change let us know. Could you also let us know what we should do to enable a future merge of this change to be successful. Is the concern around previous versions of `links.json` not being compatible with the new version if they don't follow the provided enum?

hannes-ucsc: No other components rely on this enum. It's just that TDR performs schema validation so if a `links.json` generated bi Ingest comes along with a legitimate `protocol_type` that fails to validate against this enum because the enum contains a mistake, the import would fail. I don't know if we should separate this change out or not until my questions above are answered.

hannes-ucsc: Let me also clarify my question

> Are there any staging areas in the wild containing links.json documents that use this schema?

Has this schema version (with the added enum) been used to validate any existing, real-world `links.json` documents?

json_schema/core/file/file_core.json
NoopDog: @hannes-ucsc it might be nice to also represent if a file source type is "internal" or "external". We could do this by making this an enum of dictionaries instead of strings, with an added property called "type". This would be of use in simplifying to users the distinction between DCP and CGM matrices.

NoopDog (commented):
Hi there, I would like to propose an alternative to making an optional field for matrix_cell_count on analysis_file.json.

Another way to do this is to make a matrix_file.json that includes analysis_file.json, basically extending it and making a new type, and then adding matrix_cell_count as a required field to matrix_file.json.  

We already have image_file, reference_file, sequence_file and it seems to me that since matrix files are one of our primary outputs, that we would do well to model them explicitly.

This would provide a location to attach other matrix-related properties in the future and allow the schema to specify how to validate them, and also enable filtering by the file type.

json_schema/module/protocol/matrix.json
hannes-ucsc: I will a separate create issue for this. It was meant as a suggestion. I should have made that more clear.

json_schema/type/file/analysis_file.json
kbergin: We can add it to our backlog to take advantage of this field. But I'm not sure how to backpopulate it to previously analyzed datasets..

json_schema/type/protocol/analysis/analysis_protocol.json
kbergin: I didn't have plans to do so, but if there are downstream features it would enable we can consider it. Would this be for both project matrices and individual matrices?

json_schema/system/links.json
mshadbolt: As this schema has not been released, no existing `links.json` documents are at this version so I don't know how they could be validated against this new version. The test data should be the only jsons that should be at, and therefore validated, against this proposed version. Does the importer validate against the latest schema versions, or the version that is contained within each json document? 
One thing that could have happened would be if a project has been exported from staging using this version for testing purposes, since these schema updates were released in staging. I am not sure whether or not this has occurred. Nothing in production should have `links.json` documents at this new proposed version.

json_schema/core/file/file_core.json
mshadbolt: I did some searching but I can't find anything about having an enum of objects in json schema. Or are you suggesting an additional field in the schema? If it is a 'nice to have' would it be possible to make a ticket on the repo with your proposal and we will prioritise it. Thanks!

mshadbolt: @hannes-ucsc I have incorporated the enums as requested in the PR I am making with all feedback from the review process (https://github.com/HumanCellAtlas/metadata-schema/pull/1381). We'll wait until test data has been reviewed and all expected changes are made in this branch before internal review, then squash and merge to minimise commits, then re-publish the schemas on staging.

json_schema/property_migrations.json
mshadbolt: we have removed this and rolled back the version of project in the inprogress feedback PR https://github.com/HumanCellAtlas/metadata-schema/pull/1381

json_schema/module/protocol/matrix.json
Wkt8: This shouldn't hold up the release - but we expect that the two fields should be included in the metadata .tsv file on the browser, and eventually the data_normalization and field would be able to be filtered by data consumers in the browser.

json_schema/type/protocol/analysis/analysis_protocol.json
hannes-ucsc: I think both, unless that's prohibitively difficult to implement. I've opened https://github.com/HumanCellAtlas/dcp2/issues/12 to track this separately.

tests/schema_test_files/system/test_fail_links.json
```diff
@@ -0,0 +1,47 @@
...
+  "describedBy": "https://schema.dev.data.humancellatlas.org/system/9.9.9/links.json",
```
hannes-ucsc, line 2: This URL does not point at a schema file. Is that intended?

hannes-ucsc (changes requested):

json_schema/module/protocol/matrix.json
hannes-ucsc: https://github.com/HumanCellAtlas/metadata-schema/pull/1371#discussion_r612724811

json_schema/core/file/file_core.json
```diff
@@ -49,6 +49,26 @@
             "type": "string",
             "user_friendly": "Checksum",
             "example": "e09a986c2e630130b1849d4bf9a94c06"
-	    }
+	    },
+        "file_source": {
```
danielsotirhos, line 53: Please add `ENA` to the list of enum values.
This request based on the addition of the new value in the upcoming batch of [contributor generated matrices](https://docs.google.com/spreadsheets/d/1m9tXswoNAEYJDVlCKqc9L6p0c7hbOwddkQxdqxXydqY/edit?ts=601bb5db#gid=0)

json_schema/module/protocol/matrix.json
```diff
@@ -0,0 +1,67 @@
...
+            "description": "The version number of the schema in major.minor.patch format.",
+            "type": "string",
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+        "data_normalization_methods": {
+            "description": "Method(s) used to normalize data in the matrix",
+            "type": "array",
+            "user_friendly": "Data normalization method(s)",
+            "items":{
```
hannes-ucsc, line 24:
```suggestion
            "items": {
```

tests/schema_test_files/system/test_fail_links.json
hannes-ucsc:
```
$ curl 'https://schema.dev.data.humancellatlas.org/system/9.9.9/links.json'
<!DOCTYPE html>
<html>
<head>
    <title>HCA Metadata Schemas</title>
</head>
<body>
<h1>HCA Metadata Schemas</h1>
<div id="listing"></div>
<script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
<script type="text/javascript">
  var S3BL_IGNORE_PATH = false;
  var EXCLUDE_FILE = 'index.html';
</script>
<script type="text/javascript" src="https://rawgit.com/rufuspollock/s3-bucket-listing/gh-pages/list.js"></script>
</body>
```

tests/schema_test_files/system/test_fail_links.json
```diff
@@ -0,0 +1,47 @@
...
+          "file_id": "8abc7580-ce81-4a81-9c2c-872fcb23b7cd"
+        },
+        {
+          "file_type": "supplementary_file",
+          "file_id": "cb377580-ce81-4a81-9c2c-872fcb23b7cd"
+        }
+      ]
+    }
+  ]
+
```
hannes-ucsc, line 46:
```suggestion
```

hannes-ucsc (changes requested):

json_schema/core/file/file_core.json
hannes-ucsc: I second that. Daniel is acting as my proxy here.

hannes-ucsc: @clairerye: "The reference of ENA in the spreadsheet was a mistake. Spreadsheet will be corrected."

json_schema/type/protocol/analysis/analysis_protocol.json
kbergin: Ok we can investigate more in that issue. Thanks!

kbergin (approved):

json_schema/module/protocol/matrix.json
hannes-ucsc: Not a deal-breaker. @clairerye: "Will be fixed before merged". I will post formatting configuration for Amnon.

tests/schema_test_files/system/test_fail_links.json
hannes-ucsc: Since this is a failure test, the URL doesn't matter.

tests/schema_test_files/system/test_fail_links.json
hannes-ucsc: See formatting fixes above.

hannes-ucsc (approved):

ruchim (approved):
all good

aherbst-broad (approved):

NoopDog (approved):

PR #1379, opened 2021-03-29
Fixed overlook of enum arrays

ESapenaVentura:
### Release notes

For `src/human_readable_json.py` script:
- L120-125: Added enum_values variable to account for the different scenarios an enum could be present
- L133: Changed the oneliner for the variable
 

### Reviews requested

- Need 1 or 2 Reviewers to approve because this is a code update
- Last Reviewer to review/approve, merge changes by Tuesday April 6th

PR #1381, opened 2021-03-31
Adjust schemas based on feedback, manually roll back project

mshadbolt:
This PR addresses comments and concerns raised in the review of #1371 

Changes are:
* Manually removed all traces of the `data_use_restrictions` field and `data_use_ontology module` from:
    * changelog
    * json browser
    * property migrations 
* Manually wound back the project type schema to 14.2.0 in `versions.json`
* Implemented suggested wording changes to:
    * `file_core.file_source.description`
    * `matrix.description`
    * `matrix.derivation_process.description`
    * `analysis_file.matrix_cell_count.guidelines` & `analysis_file.matrix_cell_count.description`
    * `project.ega_accessions.description`
 
 * added to the enum for the `file_core.file_source`

* For `src/human_readable_json.py` script:
   - L120-125: Added enum_values variable to account for the different scenarios an enum could be present
   - L133: Changed the oneliner for the variable

This last update was tested here https://github.com/HumanCellAtlas/metadata-schema/pull/1379 but we decided to push everything as one PR

I did not fill out the update log as I wasn't sure if we want to increment the patch version for these changes or leave them as is, i.e. we determine that our initial release to staging is a 'soft' release that we can iterate on and update based on feedback before making a 'final' release on the same versions? (If we decide this is ok we should probably add it to the spec)

After this PR is reviewed I believe we will need to delete the following released staging schemas from https://schema.staging.data.humancellatlas.org/:
* 2021-03-16T14:35:35.000Z        5.8 kB         type/project/14.2.0/project
* 2021-03-24T11:09:02.000Z        6.1 kB         type/project/14.3.0/project
* 2021-03-25T15:03:15.000Z        5.8 kB         type/project/15.0.0/project
* 2021-03-24T11:08:54.000Z        2.0 kB         module/ontology/1.0.0/data_use_ontology
* 2021-03-24T13:35:32.000Z        2.7 kB         module/protocol/1.0.0/matrix
* 2021-03-24T13:35:34.000Z        1.8 kB         type/file/6.3.0/analysis_file
* 2021-03-24T13:35:23.000Z        2.1 kB         core/file/6.2.0/file_core

We would then need to re-trigger the publisher so the updated versions of these schemas are published. (apart from the data_use_ontology which has been deleted).

### Test data changes required
* ensure the test data refers to the version 14.2.0 project type schema.
* Ensure test data has correct enum values from the revised `file_core.file_source` enum

Then we would be ready to merge and for request re-review. 

I would suggest squash and merge to minimise commits.


mshadbolt:
I am wondering whether we should also revert the changes to the `links.json` module as Hannes has raised some issues with it.

aaclan-ebi (approved):

ESapenaVentura (approved):
LGTM, great job 👍

ESapenaVentura:
> I am wondering whether we should also revert the changes to the `links.json` module as Hannes has raised some issues with it.

I have read that comment, I have a question @aaclan-ebi : Do we validate links.json? since they are generated at export, I wouldn't think so, but I am happy to be proven wrong.

json_schema/core/file/file_core.json
```diff
@@ -51,15 +51,24 @@
...
+                "Contributor",
+                "ArrayExpress",
+                "HCA Release",
+                "GEO",
+                "SCEA",
+                "SCP",
+                "DCP/1 Matrix Service",
+                "LungMAP",
+                "Zenodo",
+                "Publication"
```
mshadbolt, line 67: @ESapenaVentura @Wkt8 are we okay with expanding this enum? Or was the idea to keep it very simple and lump multiple categories into the two original ones (DCP/Contributor)

lauraclarke: Out of interest, what is the difference between DCP and HCA release?

mshadbolt: I think `HCA release` refers to the files that were created as part of the `r-`release that was never really released

lauraclarke: Okay, it was my understanding those files shouldn't be released ever

mshadbolt: it was my understanding that you said you wanted to release the cell type annotations that were produced as part of that release. I.e. the cell type files that provided matching between barcodes, authors cell types and harmonised ontology terms

lauraclarke: At this point in time I worry that presenting those alongside everything might get confusing

mshadbolt: ok I remember checking with you about this at the time they were added, but can request they are removed.

Wkt8: The idea was to keep it simple and see if we wanted to expand it in the future, but I don't see any real reason not to expand it if it's been requested!

PR #1388, opened 2021-05-06
Added estimated_cell_count field to project. Fixes #1337

ESapenaVentura:
### Release notes

For `type/project/project` schema:
- Added non-required field "estimated_cell_count" HumanCellAtlas/metadata-schema#1337

PR #1396, opened 2021-07-07
Added HCA publication status to publications module. Fixes #1345

### LINKED ISSUE #1345: add hca_publication flag to publications module

publications.json

Add an optional field to indicate when a publication has been through official process to be classified as an HCA publication (https://www.humancellatlas.org/publications/)

* Field name: hca_publication_status
* Field description: Has the publication been accepted as an official HCA publication according to the process at https://www.humancellatlas.org/publications/
* Field type: boolean
* Required: no

We don't currently have a way of identifying which of the projects in the browser are official HCA publications.

ESapenaVentura:
### Release notes

For `module/project/publication` schema:
- Added non-required field `is_hca_publication` #1345

### DIFF
--- a/docs/jsonBrowser/module.md
+++ b/docs/jsonBrowser/module.md
@@ -365,7 +365,7 @@ title | The title of the publication. | string | yes |  | Publication title |  |
 doi | The publication digital object identifier (doi) of the publication. | string | no |  | Publication DOI |  | 10.1016/j.cell.2016.07.054
 pmid | The PubMed ID of the publication. | integer | no |  | Publication PMID |  | 27565351
 url | A URL for the publication. | string | no |  | Publication URL |  | https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/
-is_hca_publication | Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ? | boolean | no |  | HCA Publication? |  | yes; no
+is_hca_publication | Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ? | boolean | yes |  | HCA Publication? |  | yes; no
 
 ## Human-specific<a name='Human-specific'></a>
 _Information specific to a donor that is a human (Homo sapiens)._
--- a/docs/jsonBrowser/module.md
+++ b/docs/jsonBrowser/module.md
@@ -365,7 +365,7 @@ title | The title of the publication. | string | yes |  | Publication title |  |
 doi | The publication digital object identifier (doi) of the publication. | string | no |  | Publication DOI |  | 10.1016/j.cell.2016.07.054
 pmid | The PubMed ID of the publication. | integer | no |  | Publication PMID |  | 27565351
 url | A URL for the publication. | string | no |  | Publication URL |  | https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/
-is_hca_publication | Has the publication been accepted as an official HCA publication, acording to the process described in https://www.humancellatlas.org/publications/ ? | boolean | no |  | HCA Publication? |  | yes; no
+is_hca_publication | Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ? | boolean | no |  | HCA Publication? |  | yes; no
 
 ## Human-specific<a name='Human-specific'></a>
 _Information specific to a donor that is a human (Homo sapiens)._
--- a/docs/jsonBrowser/module.md
+++ b/docs/jsonBrowser/module.md
@@ -365,6 +365,7 @@ title | The title of the publication. | string | yes |  | Publication title |  |
 doi | The publication digital object identifier (doi) of the publication. | string | no |  | Publication DOI |  | 10.1016/j.cell.2016.07.054
 pmid | The PubMed ID of the publication. | integer | no |  | Publication PMID |  | 27565351
 url | A URL for the publication. | string | no |  | Publication URL |  | https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/
+is_hca_publication | Has the publication been accepted as an official HCA publication, acording to the process described in https://www.humancellatlas.org/publications/ ? | boolean | no |  | HCA Publication? |  | yes; no
 
 ## Human-specific<a name='Human-specific'></a>
 _Information specific to a donor that is a human (Homo sapiens)._
--- a/docs/jsonBrowser/module.md
+++ b/docs/jsonBrowser/module.md
@@ -365,7 +365,7 @@ title | The title of the publication. | string | yes |  | Publication title |  |
 doi | The publication digital object identifier (doi) of the publication. | string | no |  | Publication DOI |  | 10.1016/j.cell.2016.07.054
 pmid | The PubMed ID of the publication. | integer | no |  | Publication PMID |  | 27565351
 url | A URL for the publication. | string | no |  | Publication URL |  | https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/
-is_hca_publication | Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ? | boolean | yes |  | HCA Publication? |  | yes; no
+official_hca_publication | Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ? | boolean | no |  | Official HCA Publication? |  | yes; no
 
 ## Human-specific<a name='Human-specific'></a>
 _Information specific to a donor that is a human (Homo sapiens)._
--- a/json_schema/module/project/publication.json
+++ b/json_schema/module/project/publication.json
@@ -54,6 +54,13 @@
             "type": "string",
             "user_friendly": "Publication URL",
             "example": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/"
+        },
+        "is_hca_publication": {
+            "description": "Has the publication been accepted as an official HCA publication, acording to the process described in https://www.humancellatlas.org/publications/ ?",
+            "type": "boolean",
+            "user_friendly": "HCA Publication?",
+            "guidelines": "Should be one of: yes, or no.",
+            "example": "yes; no"
         }
     }
 }
--- a/json_schema/module/project/publication.json
+++ b/json_schema/module/project/publication.json
@@ -56,7 +56,7 @@
             "example": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/"
         },
         "is_hca_publication": {
-            "description": "Has the publication been accepted as an official HCA publication, acording to the process described in https://www.humancellatlas.org/publications/ ?",
+            "description": "Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ?",
             "type": "boolean",
             "user_friendly": "HCA Publication?",
             "guidelines": "Should be one of: yes, or no.",
--- a/json_schema/module/project/publication.json
+++ b/json_schema/module/project/publication.json
@@ -4,7 +4,8 @@
     "additionalProperties": false,
     "required": [
         "authors",
-        "title"
+        "title",
+        "is_hca_publication"
     ],
     "title": "Publication",
     "name": "publication",
--- a/json_schema/module/project/publication.json
+++ b/json_schema/module/project/publication.json
@@ -56,10 +56,10 @@
             "user_friendly": "Publication URL",
             "example": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/"
         },
-        "is_hca_publication": {
+        "official_hca_publication": {
             "description": "Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ?",
             "type": "boolean",
-            "user_friendly": "HCA Publication?",
+            "user_friendly": "Official HCA Publication?",
             "guidelines": "Should be one of: yes, or no.",
             "example": "yes; no"
         }
--- a/json_schema/module/project/publication.json
+++ b/json_schema/module/project/publication.json
@@ -5,7 +5,7 @@
     "required": [
         "authors",
         "title",
-        "is_hca_publication"
+        "official_hca_publication"
     ],
     "title": "Publication",
     "name": "publication",
@@ -59,7 +59,7 @@
         "official_hca_publication": {
             "description": "Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ?",
             "type": "boolean",
-            "user_friendly": "Official HCA Publication?",
+            "user_friendly": "Official HCA Publication",
             "guidelines": "Should be one of: yes, or no.",
             "example": "yes; no"
         }
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,18 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
 
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/develop)
 
+### [module/project/publication.json - v6.1.0] - 2021-08-13
+### Added
+Added required official_hca_publication field. Fixes #1345
+
+### [type/biomaterial/cell_line.json - v14.6.0] - 2021-08-13
+### Added
+Added required official_hca_publication field. Fixes #1345
+
+### [type/project/project.json - v14.4.0] - 2021-08-13
+### Added
+Added required official_hca_publication field. Fixes #1345
+
 ## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
 
 ### [type/project/project.json - v14.3.0] - 2021-07-26
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,12 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
 
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/develop)
 
+## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
+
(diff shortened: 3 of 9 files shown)


ESapenaVentura:
**Note**: I have left the field as non-required for now, there was a strong feeling to make it required but I think we didn't have enough time to discuss at the meeting.

json_schema/module/project/publication.json
```diff
@@ -54,6 +54,13 @@
             "type": "string",
             "user_friendly": "Publication URL",
             "example": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/"
+        },
+        "is_hca_publication": {
```
ruchim, line 58: is it likely/possible that a project has more than 1 publication?

ruchim (changes requested):
I would love one question answered before approval. 

How is this publication information meant to be used in the HCA data browser? If we can add this in the description of the PR -- it may simplify future reviews.

json_schema/module/project/publication.json
ESapenaVentura: Yes, it is possible, although if I have to go by gut intuition, not very likely. This module gets loaded in 

https://github.com/HumanCellAtlas/metadata-schema/blob/c1e477d50f64c65339fa1f8c83b29aff2a753aed/json_schema/type/project/project.json#L61-L67

as an array, so multiple publications is possible. I can't remember if there is currently any case in the DCP but we have recorded instances of the same project in more than one publication in our backoffice.

ESapenaVentura:
> I would love one question answered before approval.
> 
> How is this publication information meant to be used in the HCA data browser? If we can add this in the description of the PR -- it may simplify future reviews.

Noted - I'll try to be more clear about end goals in the future!

For this scenario, this flag may be used to highlight the official HCA publications (Selected by the HCA scientific committee) in the browser. Currently we have no way of highlighting these, and they are supposed to be of high impact to the scientific goal of the HCA. @clairerye may have a little more information on how we would like this to be displayed.

docs/jsonBrowser/module.md
```diff
@@ -365,6 +365,7 @@ title | The title of the publication. | string | yes |  | Publication title |  |
 doi | The publication digital object identifier (doi) of the publication. | string | no |  | Publication DOI |  | 10.1016/j.cell.2016.07.054
 pmid | The PubMed ID of the publication. | integer | no |  | Publication PMID |  | 27565351
 url | A URL for the publication. | string | no |  | Publication URL |  | https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/
+is_hca_publication | Has the publication been accepted as an official HCA publication, acording to the process described in https://www.humancellatlas.org/publications/ ? | boolean | no |  | HCA Publication? |  | yes; no
```
aherbst-broad, line 368: Typo: 
```suggestion
is_hca_publication | Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ? | boolean | no |  | HCA Publication? |  | yes; no
```

json_schema/module/project/publication.json
```diff
@@ -54,6 +54,13 @@
...
+            "description": "Has the publication been accepted as an official HCA publication, acording to the process described in https://www.humancellatlas.org/publications/ ?",
+            "type": "boolean",
+            "user_friendly": "HCA Publication?",
+            "guidelines": "Should be one of: yes, or no.",
+            "example": "yes; no"
```
aherbst-broad, line 63: I believe boolean properties can only be one of `true` or `false`

json_schema/module/project/publication.json
```diff
@@ -54,6 +54,13 @@
             "type": "string",
             "user_friendly": "Publication URL",
             "example": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/"
+        },
+        "is_hca_publication": {
+            "description": "Has the publication been accepted as an official HCA publication, acording to the process described in https://www.humancellatlas.org/publications/ ?",
```
aherbst-broad, line 59: Typo:
```suggestion
            "description": "Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ?",
```

aherbst-broad (commented):
Left a few questions.

json_schema/module/project/publication.json
ruchim: awesome -- so there's two different things happening here -- one which is the property name implies the response is true/false (has_*) and then we want an array list to consume hca publications. I might propose that we just call this property "hca_publication" (so the boolean assumption no longer applies) and then the array can be 0 or n elements

ruchim:
> > I would love one question answered before approval.
> > How is this publication information meant to be used in the HCA data browser? If we can add this in the description of the PR -- it may simplify future reviews.
> 
> Noted - I'll try to be more clear about end goals in the future!
> 
> For this scenario, this flag may be used to highlight the official HCA publications (Selected by the HCA scientific committee) in the browser. Currently we have no way of highlighting these, and they are supposed to be of high impact to the scientific goal of the HCA. @clairerye may have a little more information on how we would like this to be displayed.

Got it -- so we are thinking this means on a project page -- we can better highlight official HCA publications. Is there a need to be able to "filter" by whether a project has HCA publications or not? or whether we expect HCA publications information as something that can be exported/downloaded when doing a faceted search?

kbergin (approved):
I second Ruchi's comment about the term implying it's boolean when it may actually be an array of publications. But none of these things will impact analysis so I'm approving from the analysis perspective.

json_schema/module/project/publication.json
hannes-ucsc: The name is a bit awkward. "publication" is implied by the context. There is no precedent for an "is_" prefix in the name of other boolean properties:

https://github.com/HumanCellAtlas/metadata-schema/search?l=JSON&q=boolean

Can I throw 

```suggestion
        "part_of_hca": {
```

in the ring? This is based on the wording on https://www.humancellatlas.org/publications/

hannes-ucsc: Alternatively, we could model this more richly by making this an enum that tracks the review process:

```
"hca_review": {
    "type": "string",
    "enum": [ "pending", "rejected", "accepted" ]
```

hannes-ucsc (changes requested):
I don't feel strongly about my comments below but I want to know what the reason is for not making this required. Generally speaking, it is better to be strict first and relax the requirements later if needed than to start lenient and later trying to tighten the requirements. The former is a backwards compatible update, the latter isn't and requires schema migration.

json_schema/module/project/publication.json
ESapenaVentura: Yes, you are right, but we currently gather the metadata from contributors in a spreadsheet. To avoid confusion for non-programmatic users, the spreadsheet importer automatically transforms "yes" and "no" into the boolean value. We have other instances of this in the schema: 
https://github.com/HumanCellAtlas/metadata-schema/blob/c1e477d50f64c65339fa1f8c83b29aff2a753aed/json_schema/module/project/contact.json#L70-L75

We display the example in the spreadsheet as help text on top to guide on what the answer should look like, so it's very useful for the contributor!

json_schema/module/project/publication.json
ESapenaVentura: Good catch!

json_schema/module/project/publication.json
ESapenaVentura: >I think there's a misunderstanding here. This schema describes a single publication. The project schema has a property that is a list of documents that match this schema. Any number of the project's publications can be an official HCA publication. I think this is modeled correctly.

Yes, spot on! The field itself is not an array, but the property in the `project` type schema that loads the module is.

About the name of the field, I've been discussing with the team, and `part_of_hca` does not sound less confusing to some members, as some members pointed out everything we ingest will be part of HCA, even if it's not an official hca publication. How about we drop the `is` for an `official`? (`official_hca_publication`?)

I understand there is a concern about redundancy, but that's how they are officially named, hca publications. Happy to jump on a call to discuss further about this!

ESapenaVentura: >Alternatively, we could model this more richly by making this an enum that tracks the review process:
>.
>.


About the last comment, I like the idea to make the model more rich, but I think it will make the process very complicated. We usually only have access to the final decision if it's positive, so getting (and updating over time when it changes!) the status may make it a bit complicated.

ESapenaVentura:
> I don't feel strongly about my comments below but I want to know what the reason is for not making this required. Generally speaking, it is better to be strict first and relax the requirements later if needed than to start lenient and later trying to tighten the requirements. The former is a backwards compatible update, the latter isn't and requires schema migration.

Agree. I don't think it will cause significant overhead (It's not a piece of metadata that is hard to get), so making it required and relaxing the requirements later if needed sounds good to me. I will apply the change!

json_schema/module/project/publication.json
NoopDog: @ESapenaVentura Yes`official_hca_publication` seems fine. Also would be great to make this required if possible. 

I presume that making this field optional for now makes it possible to do a simpler data migration and mark the official HCA papers only. Is this correct? Or is it about the same effort to then mark all non HCA publications "false" in the same go? 

My preference would be to make it required now if it does not create an undue burden on the data ops to backfill. 

My understanding (and correct me if I am wrong) is that the metadata is "Open world" meaning that the absence of a fact does not imply the fact is false, but only that it is unknown by the system.

aherbst-broad (approved):

json_schema/module/project/publication.json
ESapenaVentura: Perfect, thanks! `official_hca_publication` it is then!

>I presume that making this field optional for now makes it possible to do a simpler data migration and mark the official HCA papers only. Is this correct? Or is it about the same effort to then mark all non HCA publications "false" in the same go?

Whenever we do a data migration, we do have (in the tracking spreadsheet) a field to indicate papers that have been accepted as HCA publications, so (even if it's not super thorough) we may already have something to mark papers as hca publications!

>My preference would be to make it required now if it does not create an undue burden on the data ops to backfill.

I changed it yesterday, it is required now! I think the biggest burden will be how to "massively" migrate the project json, but I think we have the tools and it's just a matter of having someone on the team work on it.

>My understanding (and correct me if I am wrong) is that the metadata is "Open world" meaning that the absence of a fact does not imply the fact is false, but only that it is unknown by the system.

💯  spot on!

ESapenaVentura:
> > > I would love one question answered before approval.
> > > How is this publication information meant to be used in the HCA data browser? If we can add this in the description of the PR -- it may simplify future reviews.
> > 
> > 
> > Noted - I'll try to be more clear about end goals in the future!
> > For this scenario, this flag may be used to highlight the official HCA publications (Selected by the HCA scientific committee) in the browser. Currently we have no way of highlighting these, and they are supposed to be of high impact to the scientific goal of the HCA. @clairerye may have a little more information on how we would like this to be displayed.
> 
> Got it -- so we are thinking this means on a project page -- we can better highlight official HCA publications. Is there a need to be able to "filter" by whether a project has HCA publications or not? or whether we expect HCA publications information as something that can be exported/downloaded when doing a faceted search?

I think, at least as a start, it would be good to display that information at the level of the project page. From there, we can work out if we need to provide with those 2 features you mention - Being able to filter projects by HCA publications and exporting/downloading publication info from those HCA publications!

json_schema/module/project/publication.json
```diff
@@ -4,7 +4,8 @@
     "additionalProperties": false,
     "required": [
         "authors",
-        "title"
+        "title",
+        "is_hca_publication"
```
NoopDog, line 8:
```suggestion
        "official_hca_publication"
```

json_schema/module/project/publication.json
```diff
@@ -54,6 +55,13 @@
             "type": "string",
             "user_friendly": "Publication URL",
             "example": "https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5667944/"
+        },
+        "official_hca_publication": {
+            "description": "Has the publication been accepted as an official HCA publication, according to the process described in https://www.humancellatlas.org/publications/ ?",
+            "type": "boolean",
+            "user_friendly": "Official HCA Publication?",
```
NoopDog, line 62:
```suggestion
            "user_friendly": "Official HCA Publication",
```

json_schema/module/project/publication.json
NoopDog: Hi @ESapenaVentura it seems to me the documentation should describe the metadata and how it will appear in the json, not how it will appear in the ingest spreadsheet. Is there a way the spreadsheet display can apply the correct transform rather than using "yes, no" here?`

NoopDog (changes requested):
I made two suggestions below to fix the name of `official_hca_publication` in the required list, to remove the "?" from the friendly name and I also left a comment about using "yes" "no" instead of "true" "false" for the example values of the `boolean` field.  I feel the documentation should document the metadata and the json, and not be concerned with the input spreadsheet.

json_schema/module/project/publication.json
ESapenaVentura: Thanks for the suggestion! Being a boolean property, I'd prefer to keep the question mark in the user friendly field, is there any issue with that?

hannes-ucsc: No.

json_schema/module/project/publication.json
hannes-ucsc: The fact that some schema properties in this repository are taylored towards spreadsheet generation is a separate issue that we're not going to fix in this PR.

hannes-ucsc (changes requested):
One show-stopping issue.

hannes-ucsc:
> > I would love one question answered before approval.
> > How is this publication information meant to be used in the HCA data browser? If we can add this in the description of the PR -- it may simplify future reviews.
> 
> Noted - I'll try to be more clear about end goals in the future!
> 
> For this scenario, this flag may be used to highlight the official HCA publications (Selected by the HCA scientific committee) in the browser. Currently we have no way of highlighting these, and they are supposed to be of high impact to the scientific goal of the HCA. @clairerye may have a little more information on how we would like this to be displayed.

If this property is to be rendered in the Data Browser, Azul needs to index it and we would like a corresponding PR against the schema test data repo that adds this property there so we can write unit tests.

clairerye:
> > > I would love one question answered before approval.
> > > How is this publication information meant to be used in the HCA data browser? If we can add this in the description of the PR -- it may simplify future reviews.
> > 
> > 
> > Noted - I'll try to be more clear about end goals in the future!
> > For this scenario, this flag may be used to highlight the official HCA publications (Selected by the HCA scientific committee) in the browser. Currently we have no way of highlighting these, and they are supposed to be of high impact to the scientific goal of the HCA. @clairerye may have a little more information on how we would like this to be displayed.
> 
> If this property is to be rendered in the Data Browser, Azul needs to index it and we would like a corresponding PR against the schema test data repo that adds this property there so we can write unit tests.

Yes please to it being rendered in the data browser. When this PR is merged we can create the test data and will let you know when that is available. Once this PR is approved, are you happy for us to merge staging to master or do you want to sign off on the test data first and therefore the staging to master PR too? Also happy to confirm this in the post demo session on Tuesday (I have put it on the agenda).

hannes-ucsc:
As with PR [1388](https://github.com/HumanCellAtlas/metadata-schema/pull/1388#issuecomment-881558737), no master PR necessary, I would say.

json_schema/module/project/publication.json
amnonkhen: Leaving `?` in.

amnonkhen: Resolving on behalf of @ESapenaVentura who is on vacation.

json_schema/module/project/publication.json
amnonkhen: Resolving on behalf of @ESapenaVentura who is on vacation.

amnonkhen (approved):

json_schema/module/project/publication.json
NoopDog: Hi @ESapenaVentura @hannes-ucsc this is the friendly name a property. Usually, property names do not contain punction. We use this friendly name in the metadata dictionary in the data portal for example and plan to eventually use it in the data browser table column names.

Also, currently in the metadata schema, it seems that other friendly names for boolean properties do not contain punctuation. See `contact.corresponding_contributor` or `death.cold_perfused` or `death.normothermic_regional_perfusion` for examples.

clairerye: I see your point @NoopDog, lets wait until Enrique is back to resolve this. I think we all agree we want to be as consistent as possible, so I am thinking Enrique has something quite specific in mind here.

hannes-ucsc: Precedent seems to be not to have question mark:

https://github.com/HumanCellAtlas/metadata-schema/search?q=%22type+boolean%22

hannes-ucsc:
If we want to follow the yet-to-be accepted SOP, reviews should be requested from all reviewers after changes have been pushed.

json_schema/module/project/publication.json
amnonkhen: Personally, I prefer to avoid the question mark. 

From reading the comments I understood that @ESapenaVentura wants the question mark 


> Thanks for the suggestion! Being a boolean property, I'd prefer to keep the question mark in the user friendly field, is there any issue with that?

and @hannes-ucsc being ok with that.

> No.

Here's a screen shot:

image

Have I misunderstood?

hannes-ucsc: I don't have an issue with the question mark because I don't think it's a big deal. I pointed out that precedence seems to be not to have a question mark in the `user_friendly` on boolean properties. Those two statements don't contradict each other

ruchim (approved):
boolean confusion cleared -- thanks!

json_schema/module/project/publication.json
amnonkhen: After discussion the question mark with @ESapenaVentura, we decided to get rid of it.

NoopDog (approved):
Nice. Thank you!

hannes-ucsc (approved):

PR #1437, opened 2021-12-09
Allow inclusion of an external DRS URI in file descriptors

aherbst-broad:
### Context

The Lungmap project has a requirement to source files from an external system that will provide DRS URIs. These files will not be importable to TDR but the DRS URIs will need to be wired through into the file descriptor so they can be surfaced downstream in the browser. This is a draft PR capturing a proposal to allow the inclusion of these external DRS URIs.

[Related DCP system spec PR](https://github.com/HumanCellAtlas/dcp2/pull/55)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -37,6 +37,12 @@ Added Read 3, Read 4 to barcode read enum. Fixes #1401
 ### Added
 Added spatial barcode field loading barcodes module
 
+## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
+
+### [system/file_descriptor.json - v2.1.0] - 2021-03-21
+### Added
+Added optional drs_uri file to file_descriptor
+
 ### [module/project/publication.json - v7.0.0] - 2021-08-13
 ### Added
 Added required official_hca_publication field. Fixes #1345
--- a/changelog.md
+++ b/changelog.md
@@ -39,10 +39,6 @@ Added spatial barcode field loading barcodes module
 
 ## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
 
-### [system/file_descriptor.json - v2.1.0] - 2021-03-21
-### Added
-Added optional drs_uri file to file_descriptor
-
 ### [module/project/publication.json - v7.0.0] - 2021-08-13
 ### Added
 Added required official_hca_publication field. Fixes #1345
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -97,7 +97,7 @@
     },
     "drs_uri": {
       "description": "Data Repository Service URI pointing at this file. If this property is absent, a data file exists in the repository containing this descriptor. If this property is present and not `null`, a data file exists in another repository at the specified URI. If this property is present but `null`, the data file is not available, but will likely become available in the future, along with an updated descriptor referencing it.",
-      "user_friendly": "DRS URI",
+      "user_friendly": "Data Repository Service URI",
       "type": ["string", "null"],
       "format": "uri",
       "example": "drs://drs.example.org/314159"
--- a/json_schema/system/file_descriptor.json
+++ b/json_schema/system/file_descriptor.json
@@ -19,7 +19,8 @@
   "properties": {
     "describedBy": {
       "description": "The URL reference to the schema.",
-      "type": "string"
+      "type": "string",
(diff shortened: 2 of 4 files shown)


json_schema/system/file_descriptor.json
```diff
@@ -94,6 +90,17 @@
...
+      "user_friendly": "DRS URI",
+      "type": "string",
+      "format": "uri",
+      "example": "drs://drs.example.org/314159"
     }
-  }
+  },
+  "anyOf": [
+    {"required": ["drs_uri"]},
+    {"required": ["content_type", "size",  "sha256", "crc32c"]}
```
aherbst-broad, line 104: This section is intended to express a mutually exclusive condition, i.e., if a `drs_uri` is present, `content_type`, `size`, `sha256`, `crc32c` should _not_ be present (and vice-versa).

JoshuaFortriede: For a mutually exclusive condition, the "oneOf" keyword should be used. With that said, I am not sure these field lists should be mutually exclusive. I can imagine implementation layers that would want the content type, size, and hashes of a file while also utilizing a drs_uri to serve the file. For instance, a data shopping cart where you can check boxes to add files to your shopping cart, and you are summing the file size to inform the user of their total end payload.

JoshuaFortriede:
For a mutually exclusive condition, the "oneOf" keyword should be used. With that said, I am not sure these field lists should be mutually exclusive. I can imagine implementation layers that would want the content type, size, and hashes of a file while also utilizing a drs_uri to serve the file. For instance, a data shopping cart where you can check boxes to add files to your shopping cart, and you are summing the file size to inform the user of their total end payload.

hannes-ucsc (changes requested):
~Tag~

aherbst-broad:
@JoshuaFortriede I believe the array notation is equivalent to `anyOf` (both enforce an `or` condition). See the example here https://cswr.github.io/JsonSchema/spec/multiple_types/

aherbst-broad:
@JoshuaFortriede does your comment still hold? The current rev of this PR is not using an `anyOf` condition for the required properties and instead has reverted to the original set of required properties.

ESapenaVentura (changes requested):
Given the description of the field, we are only expecting a single value at a time (or the absence of the field) and therefore I don't think we need the "oneOf" keyword:
https://json-schema.org/understanding-json-schema/reference/type.html

The schema change itself looks good to me, but all the PRs must go through staging first as stated in the [commiters document](https://github.com/HumanCellAtlas/metadata-schema/blob/master/docs/committers.md#specific-how-to-for-making-changes)

hannes-ucsc:
@ESapenaVentura, we wanted to make sure that people don't accidentally create phantom files by omitting the property. That's why the PR distinguishes between absence and `null` value.

json_schema/system/file_descriptor.json
```diff
@@ -19,8 +19,7 @@
   "properties": {
     "describedBy": {
       "description": "The URL reference to the schema.",
-      "type": "string",
-      "pattern" : "^https?://schema.(.*)?humancellatlas.org/system/(([0-9]{1,}.[0-9]{1,}.[0-9]{1,})|([a-zA-Z]*?))/file_descriptor$"
```
NoopDog, line 23: Curious why `pattern` is being deleted here.

aherbst-broad: @NoopDog good catch, that was removed to allow the lungmap folks to test and develop while this schema change was in flight. It has been restored.

aherbst-broad:
@ESapenaVentura apologies for not following the SOP, I've rebased against the staging branch according to the committers doc.

hannes-ucsc (changes requested):
Conflicts.

changelog.md
```diff
@@ -41,6 +41,8 @@ Added Read 3, Read 4 to barcode read enum. Fixes #1401
 ### Added
 Added spatial barcode field loading barcodes module
 
+## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
+
```
hannes-ucsc, line 45: Don't know why this change is here. It seems unrelated but I don't have sufficient insight into the release process use in this repository.

hannes-ucsc (approved):

hannes-ucsc:
Continued in https://github.com/HumanCellAtlas/metadata-schema/pull/1575

PR #1454, opened 2022-03-08
Staging Release 2022-08-03

### LINKED ISSUE #1441: Make Imaging fields non-required: numerical aperture, pixel size and overlapping tiles

- `type/protocol/imaging/imaging_protocol`

I would like to make those fields optional

In order to be able to ingest certain technologies, and based on the discussions @gabsie and @ami-day had with some scientists, those fields are not essential and therefore should not be required to ingest imaging datasets

ESapenaVentura:
### Release notes

For /type/protocol/imaging/imaging_protocol schema:

- Changed overlapping_tiles, numerical_aperture and pixel_size non-required. Fixes #1441

### Versions:
json_schema/type/protocol/imaging/imaging_protocol: 11.4.0


Wkt8 (approved):
LGTM!

PR #1462, opened 2022-05-13
 Expand disease ontology to include phenotype - Fixes #1460 

### LINKED ISSUE #1460: Expand disease ontology to include phenotype

I would like to request an update to the `module/ontology/disease_ontology.json` schema.

Currently, we are restricted to ontology terms which are a from:
- obo:mondo which are a subclass of MONDO:0000001 (Mondo Disease Ontology)
- obo:efo which are a subclass of PATO:0000461 (Normal)

We would like to add a new branch for phenotypic observations which are not strictly a disease but have clinical relevance and are of interest for reaserch.
We should expand the ontology term restrictions for disease_ontology to include 'phenotype' (exact branch tbd - it shouldn't be just human phenotype?).

``` 
"graph_restriction":  {
                "ontologies" : ["obo:mondo", "obo:efo"],
                "classes": ["MONDO:0000001", "PATO:0000461"],`
```

This change will allow contributors to report clinically signigcant phenotype observation in a way that is searchable

idazucchi:
Schema update priority ticket with explanation on the change: https://github.com/HumanCellAtlas/metadata-schema/issues/1460

### Release notes

For module/ontology/disease_ontology schema:
- changed restrictions to include `phenotype` branch in the allowed terms.
 

### Reviews requested

- This is a minor update
- Last Reviewer to review/approve, merge changes by 27/05/2022

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,22 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
 
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/staging)
 
+### [module/ontology/disease_ontology.json - v5.4.0] - 2022-05-24
+### Added
+Added the phenotype branch to graphRestrictions. Fixes #1460
+
+### [type/biomaterial/cell_line.json - v15.1.0] - 2022-05-24
+### Added
+Added the phenotype branch to graphRestrictions. Fixes #1460
+
+### [type/biomaterial/donor_organism.json - v15.6.0] - 2022-05-24
+### Added
+Added the phenotype branch to graphRestrictions. Fixes #1460
+
+### [type/biomaterial/specimen_from_organism.json - v10.5.0] - 2022-05-24
+### Added
+Added the phenotype branch to graphRestrictions. Fixes #1460
+
 ## [Released](https://github.com/HumanCellAtlas/metadata-schema/)
 
 ### [module/ontology/file_content_ontology.json - v1.1.0] - 2022-03-24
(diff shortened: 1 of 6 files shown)


ESapenaVentura (approved):
LGTM 👍

hannes-ucsc (approved):

ncalvanese1:
This all looks good to me from the importer side!

ruchim:
does it still make sense for the category to be called "disease" if the values in that category are "disease" + "phenotype"? It sounds to me like phenotype is the more generic term, and disease is the more specific one. If that line of thought makes sense, we'd want to re-name this filter in the UI to say "phenotype". I'm not a data modeling expert, so please take this with a grain of salt/skepticism.

ruchim (approved):
The changes look technically right. I'd love if we can continue the discussion around how this expanded definition of accepted ontologies may impact the researcher experience at the HCA portal and if its worthwhile to re-classify any filters.

amnonkhen (approved):

idazucchi:
@ruchim 
We discussed the possibility of renaming the disease field between the wrangler, and ultimately we agreed that it still makes sense to keep the name disease. Our intent with the phenotype ontology is to represent conditions that indicate a deviation from the normal, healthy status. For example we might want to use the term Liver fibrosis, which indicates the presence of scar tissue in the liver. On its own phenotype is an extremely broad term which doesn’t necessarily refer to unhealthy status, so it would not be indicative of what can be found in that field

willrockout:
@idazucchi @ESapenaVentura I think this update may have broken something in ingest related to ontology lookups. After fixing a linking issue with this projects submission 135f7f5c-4a85-4bcf-9f7c-4f035ff1e428 we've started getting the below error for a number of biomaterials:
Screen Shot 2022-05-24 at 12 06 46 PM

PR #1473, opened 2022-07-06
Added permeabilisation time field with time units. Fixes #1471

### LINKED ISSUE #1471: Permeabilisation time and permeabilisation time unit in imaging_preparation_protocol
I would like to request an update to the `imaging_preparation_protocol` schema.

I would like to add two new optional fields, `permeabilisation_time` and `permeabilisation_time_unit` to the imaging_preparation_protocol schema 

'permeabilisation_time_unit' would be validated against the ontology module.

**What new field(s) need to be changed/added?
* Field name: permeabilisation_time
* Field description: The permeabilisation time in permeabilisation time units that the tissues were permeabilised for 
* Field type: number
* Required: no
* Examples: 15, 20-25
* Guidelines: "Enter either a single value or a range of values. Indicate a range using a hyphen."

Field 2
* Field name: permeabilisation_time_unit
* Field description: The unit in which permeabilisation time is expressed
* Field type: object
* Required: no
* Examples: 42
* CV or enum: "module/ontology/time_unit_ontology.json",

Data contributors need to be able to store and find permeabilisation_time as this is a key feature that they are interested in Visium experiments.

ESapenaVentura:
### Release notes

For `type/protocol/imaging/imaging_preparation_protocol.json` schema:
- Added non-required field `permeabilisation_time`
- Added non-required field `permeabilisation_time_unit`
- Added dependency between permeabilisation_time and permeabilisation_time_unit
 

### Reviews requested

- This is a minor update
- Reviews are requested as per usual

### Rationale
Based on the findings by the education team, exposed in the [May 24th demos](https://docs.google.com/presentation/d/10-Ei7SXCqyrLLlWyAGWxS0E02bG6r-sIAumyHBtDoIo/edit#slide=id.g12e39c13645_1_0), we have decided to add the "permeabilisation time" field to the imaging_preparation_protocol to ensure we are capturing the information needed to represent Visium experiments

see #1471

### DIFF
--- a/docs/jsonBrowser/type.md
+++ b/docs/jsonBrowser/type.md
@@ -255,6 +255,8 @@ post_final_slicing_interval | Length of time between secondary slicing and hybri
 post_final_slicing_interval_unit | The unit of time in which the post final slicing interval is expressed. | object | no | [See module  time_unit_ontology](module.md/#time_unit_ontology) | Post final slicing interval time unit |  | day
 fiducial_marker | Fiducial markers for the alignment of images taken across multiple rounds of imaging. | string | no |  | Fiducial marker |  | beads
 expansion_factor | Factor by which the imaged tissue was expanded in one dimension. | number | no |  | Expansion factor |  | 3
+permeabilisation_time | The permeabilisation time in time units that the tissue was exposed to. | number | no |  | Permeabilisation time |  | 12
+permeabilisation_time_unit | The unit in which Permeabilisation time is expressed. | object | no | [See module  time_unit_ontology](module.md/#time_unit_ontology) | Permeabilisation time unit |  | 
 
 ## Imaging Protocol
 _Information about the imaging protocol._
--- a/docs/jsonBrowser/type.md
+++ b/docs/jsonBrowser/type.md
@@ -256,7 +256,7 @@ post_final_slicing_interval_unit | The unit of time in which the post final slic
 fiducial_marker | Fiducial markers for the alignment of images taken across multiple rounds of imaging. | string | no |  | Fiducial marker |  | beads
 expansion_factor | Factor by which the imaged tissue was expanded in one dimension. | number | no |  | Expansion factor |  | 3
 permeabilisation_time | The permeabilisation time in time units that the tissue was exposed to. | number | no |  | Permeabilisation time |  | 12
-permeabilisation_time_unit | The unit in which Permeabilisation time is expressed. | object | no | [See module  time_unit_ontology](module.md/#time_unit_ontology) | Permeabilisation time unit |  | 
+permeabilisation_time_unit | The unit in which permeabilisation time is expressed. | object | no | [See module  time_unit_ontology](module.md/#time_unit_ontology) | Permeabilisation time unit |  | 
 
 ## Imaging Protocol
 _Information about the imaging protocol._
--- a/json_schema/type/protocol/imaging/imaging_preparation_protocol.json
+++ b/json_schema/type/protocol/imaging/imaging_preparation_protocol.json
@@ -7,6 +7,9 @@
         "schema_type",
         "protocol_core"
     ],
+    "dependentRequired": {
+        "permeabilisation_time": ["permeabilisation_time_unit"]
+    },
     "title": "Imaging preparation protocol",
     "name": "imaging_preparation_protocol",
     "type": "object",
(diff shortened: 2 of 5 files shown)


ncalvanese1 (approved):
Looks good to me!

NoopDog (approved):

docs/jsonBrowser/type.md
```diff
@@ -255,6 +255,8 @@ post_final_slicing_interval | Length of time between secondary slicing and hybri
 post_final_slicing_interval_unit | The unit of time in which the post final slicing interval is expressed. | object | no | [See module  time_unit_ontology](module.md/#time_unit_ontology) | Post final slicing interval time unit |  | day
 fiducial_marker | Fiducial markers for the alignment of images taken across multiple rounds of imaging. | string | no |  | Fiducial marker |  | beads
 expansion_factor | Factor by which the imaged tissue was expanded in one dimension. | number | no |  | Expansion factor |  | 3
+permeabilisation_time | The permeabilisation time in time units that the tissue was exposed to. | number | no |  | Permeabilisation time |  | 12
+permeabilisation_time_unit | The unit in which Permeabilisation time is expressed. | object | no | [See module  time_unit_ontology](module.md/#time_unit_ontology) | Permeabilisation time unit |  |
```
hannes-ucsc, line 259:
```suggestion
permeabilisation_time_unit | The unit in which permeabilisation time is expressed. | object | no | [See module  time_unit_ontology](module.md/#time_unit_ontology) | Permeabilisation time unit |  | 
```

json_schema/type/protocol/imaging/imaging_preparation_protocol.json
```diff
@@ -7,6 +7,10 @@
         "schema_type",
         "protocol_core"
     ],
+    "dependentRequired": {
+        "permeabilisation_time": ["permeabilisation_time_unit"],
+        "permeabilisation_time_unit": ["permeabilisation_time"]
+    },
```
hannes-ucsc, line 13: `dependentRequired` was [introduced](https://json-schema.org/draft/2019-09/release-notes.html#validation-vocabulary) in release 2019-09 of JSONSchema while all schemas in this repository consistently use draft 7. At best, this constraint would be ignored by the validators currently used, in the worst case it would break validation of otherwise valid documents. I don't actually know what happens. Have you tested this with a draft 7 validator?

Aside from that, it might be mode elegant to represent "measured values" (values with a unit), as nested `object`s with two required properties. That way you get to reuse measured values as, I think, what your team calls modules.

hannes-ucsc (changes requested):

json_schema/type/protocol/imaging/imaging_preparation_protocol.json
ESapenaVentura: Apologies, I didn't realise it was introduced in draft 09. You are right, at least for now we are not prepared for that!

I like your solution, but i think it needs some discussion (I am thinking, for example, how to avoid the case in which a time-related field would be filled with weight units)

I have created a ticket to keep this in our backlog/discuss [here](https://github.com/HumanCellAtlas/metadata-schema/issues/1479), and for now I will resort to our usual implementation (Which is the same as it is now, but without the `dependentRequired` keyword)

docs/jsonBrowser/type.md
ESapenaVentura: Corrected!

hannes-ucsc (approved):

json_schema/type/protocol/imaging/imaging_preparation_protocol.json
idazucchi: For draft 2019-09 `dependentRequired` substitutes in part `dependencies` from draft 7. We could use `dependencies` where Enrique initially proposed to use `dependentRequired`. 
It would be the same draft as the rest of the schema, it would be validated by the validators and it's the built-in mechanism to indicate dependencies between different parts of the schema.

hannes-ucsc: Works for me.

ESapenaVentura:
Applied the `dependencies` keyword as discussed above - Please review!

ncalvanese1 (approved):
Looks good to me!

NoopDog (approved):

amnonkhen (approved):

hannes-ucsc (approved):

PR #1513, opened 2023-02-22
Added timecourse to remaining biomatierals. Fixes #1511

### LINKED ISSUE #1511: Add timecourse module to specimen from organism, imaged specimen, organoid biomaterials

I would like to request updates to:

specimen_from_organism.json
imaged_specimen.json
organoid.json

*e.g.*: I would like to request an update to the `cell_line.json` schema.

Requesting that "module/biomaterial/timecourse.json" be added to each of the above biomaterials

*e.g.*: I would like to add a new field - `random_integer` - to this schema to allow data contributors to assign a random integer to the cell line.

Timecourse metadata may be relevant to certain specimens/organoids.

rachadele:
### Release notes

For `type/biomaterial/specimen_from_organism.json` schema:
-Added timecourse module

For `type/biomaterial/imaged_specimen.json` schema:
-Added timecourse module

For `type/biomaterial/organoid.json` schema:
-Added timecourse module

### Reviews requested

-Minor change, adding timecourse to specimen, organoid and imaged specimen biomaterials. All DCP2 reviewers need to review


NoopDog:
Hi @willrockout If we are adding this to all biomaterials, why not add to [biomaterial core](https://data.humancellatlas.org/metadata/dictionary/biomaterial/biomaterial_core) instead?
Cheers,
D

willrockout:
@NoopDog I was a bit torn on that too. I think because it's a module I was thinking of separating them. @ESapenaVentura thoughts? Not sure if changing the biomaterial core would have any other effects.

willrockout:
@NoopDog My reasoning was that the core entities were designed to be fields that are always filled out such as id, name, and taxon id. In the case of timecourse, the reason it's a module is that not every experimental design will have that component. @tburdett correct me if I'm wrong about the original intent of core, type, and module.

ncalvanese1 (approved):
Looks good to me!

hannes-ucsc (commented):
The core schema already has several optional properties so optionality shouldn't stop us from putting timecourse into biomaterial core. Also, it can remain a module and still be referenced in the core schema. There's no contradiction there either. 

Think what would happen if we added a new, concrete biomaterial type. Is it more likely that we'd want to support timecourse in that new type or not? If so, timecourse should go into the core schema. The counter argument to putting everything in the core schema (and making it optional there) is that it makes the core schema larger and harder to understand. If you already know there will be biomaterial types for which timecourse categorically makes no sense, then I wouldn't put it in the core schema, but instead in each concrete schema where I know timecourse makes sense. I don't know enough about the biology to decide that. Last thought: Another argument for leaving a property in the concrete (`type/`) schemas is if doing so allows you to make it required in some of them (thereby adding specificity).

willrockout:
I think what @hannes-ucsc says makes sense. I can't think of any examples where you'd have a biomaterial that couldn't possibly have a timecourse component. The last thing I want to check before fixing this PR is backward compatibility and the new spreadsheet creation. @amnonkhen @ESapenaVentura Would this cause any issues with projects that have timecourse filled out since this would change programmatic names?

willrockout:
Alright, I have changed the timecourse module to be in the biomaterial core and removed it from the individual biomaterial types. @ncalvanese1 @hannes-ucsc @NoopDog @ESapenaVentura @amnonkhen Can you please review again?

hannes-ucsc (approved):

NoopDog (approved):

json_schema/update_log.csv
```diff
@@ -1 +1,2 @@
 Schema,Change type,Change message,Version,Date
+core/biomaterial/biomaterial_core,minor,Added timecourse module. Fixes #1511,,
```
ESapenaVentura, line 2: We also need the lines for the biomaterials that lost the "timecourse" property (Major change, since a property was deleted)

ESapenaVentura (changes requested):
The changes to the schemas LGTM, just 2 comments:

- Missing lines in update_log.csv (See comment)
- property_migrations.json needs to be updated with the "moved" property (Happy to help with that if you're unsure how to update it)

willrockout:
@NoopDog @ESapenaVentura @hannes-ucsc @ncalvanese1 @amnonkhen Final updates have happened if you can review them one more time?

ncalvanese1 (approved):

ESapenaVentura (approved):
LGTM, just a note on internal processes, the script that updates the log/the versions should be run right before merging.

This is to allow for multiple PRs to be operating at the same time, as if another PR closes earlier and bumps the versions of the same entities, you can bump over them right before merging!

I think in this case it should be ok because we don't have anything going on for biomaterials at the moment, but just so you know!

hannes-ucsc (approved):

NoopDog (approved):

amnonkhen (approved):

json_schema/update_log.csv
rachadele, line 1: why was this deleted?

willrockout:
Fixed schema versions issues so no need for re-review.

changelog.md
```diff
@@ -7,7 +7,66 @@ and (starting with v4.0.0) this project adheres to [Semantic Versioning](http://
 
 ## [Unreleased](https://github.com/HumanCellAtlas/metadata-schema/tree/staging)
 
+<<<<<<< HEAD
```
hannes-ucsc, line 10: These conflict markers should not be here.

hannes-ucsc (changes requested):
> Update fixed schema versions issues so no need for re-review.

The SOP requires re-review after every change. Not every small change is trivial. If the changes are trivial, then reviewing them can be done in a minute.

willrockout:
This has been moved to PR #1529.

Issue #1517, opened 2023-03-15
Update project tab with four metadata entities for grouping Tissue Atlas projects


Wkt8:
### For which schema is a change/update being suggested?
Create a new 'hca_bionetwork' module and add it to project.json

### What should the change/update be?
Creating four fields as part of the 'hca_bionetwork' module and adding them to project.json

#### 1. HCA Bionetworks
I would like to add a new field - `Project.hca_bionetwork.name` - to this schema. This will allow data contributors to denote if the project is part of a HCA bionetwork and if so, which one.

This update constitutes a minor change to the schema(s) it affects.

```
Field name: project.hca_bionetwork.name
Field description: A field describing if the project is part of an HCA Bionetwork (e.g. Kidney). Enter ‘No’ if the project is not part of an HCA Bionetwork
Field type: array
Required: no
Examples: Kidney; Lung; No
CV or enum: enum
```

#### 2. HCA Tissue Atlas
I would like to add a new field - `Project.hca_bionetwork.hca_tissue_atlas` - to this schema. This will allow data contributors to denote if the project is part of a HCA Tissue Atlas and if so, which one. In this case, a Tissue Atlas represents a singular h5ad file. 

This update constitutes a minor change to the schema(s) it affects.

```
Field name: project.hca_bionetwork.hca_tissue_atlas
Field description: A field describing if the project is part of a HCA Tissue Atlas (e.g. Kidney). 
Field type: string, array
Required: no
Examples: Brain Alzheimer's Atlas 
CV or enum: Free Text
```

#### 3. HCA Tissue Atlas Version
I would like to add a new field - `Project.hca_bionetwork.hca_tissue_atlas_version` - to this schema. This will allow data contributors to denote what version of the Tissue Atlas is associated with this publication (e.g. v1.0; v2.0)

This update constitutes a minor change to the schema(s) it affects.

```
Field name: project.hca_bionetwork.hca_tissue_atlas_version
Field description: A field describing the version of the HCA Tissue Atlas is associated with the publication (e.g. v1.0; v2.0)
Field type: string, array
Required: no
Examples: 'v1.0; v2.0'
CV or enum: REGEX
```

#### 2. Integrated project
I would like to add a new field - `Project.hca_bionetwork.atlas_project` - to this schema. This will allow data contributors to denote if the publication is the HCA Tissue Atlas publication which integrates data from other projects within the atlas.

This update constitutes a minor change to the schema(s) it affects.

```
[...]

PR #1526, opened 2023-06-29
dcp-954 fix hca_bionetwork schema url

amnonkhen:
schema url (`describedBy`) pattern was incorrect.

ebi-ait/dcp-ingest-central#954
Fixes: #1527
Related to: #1517


ncalvanese1 (approved):
LGTM!

ESapenaVentura (changes requested):
It's missing writing the changes in the update_log.csv - Also not sure if the human-readable markdowns have been generated 

Here is a document with the specifics on what is expected that is done during an update https://github.com/HumanCellAtlas/metadata-schema/blob/master/docs/committers.md#specific-how-to-for-making-changes, but we can discuss on monday if it's unclear!

hannes-ucsc (changes requested):
Title is misleading. This cannot be an expedited review. See section 1.6.1 of the SOP.

NoopDog (approved):
Nice ! ⭐

hannes-ucsc (approved):

PR #1532, opened 2023-09-04
Added transplant field to specimen_from_organism. Fixes #1531

### LINKED ISSUE #1531: Add transplant field to specimen

I'd like to add a `transplant` field to clearly indicate if the sampled tissue comes from a transplanted organ

<!-- For each new field requested, please provide the following:
*e.g.*:
* Field name: transplant
* Field description: This field describes tissues sampled from transplanted organs
* Field type: string
* Required: no
* Examples: tissue from transplanted organ
* CV or enum: ["tissue from transplanted organ", "xenograft tissue"]
* Guidelines: Should be one of: tissue from transplanted organ, xenograft tissue

The new field is needed to clearly indicate whether the sampled tissue comes from a transplanted organ

idazucchi:
### Release notes

For `specimen_from_organism.json` schema:
- added `transplant` field

### Why are these changes needed?
 The new field is needed to clearly indicate whether the sampled tissue comes from a transplanted organ. This field was requested by the kidney Bionetwork.

### Reviews requested
This is a minor schema changes, all DCP2 reviewers need to review

### DIFF
--- a/json_schema/type/biomaterial/specimen_from_organism.json
+++ b/json_schema/type/biomaterial/specimen_from_organism.json
@@ -83,6 +83,17 @@
             "user_friendly": "Adjacent disease(s)",
             "guidelines": "If a healthy specimen is sampled from a site adjacent to diseased tissue, enter that tissue's disease here. If no diseased tissue is adjacent to the specimen, leave blank."
         },
+        "transplant": {
+            "description": "Description of tissue sampled from a transplanted organ.",
+            "type": "string",
+            "enum": [
+                "tissue from transplanted organ",
+                "xenograft tissue"
+            ],
+            "user_friendly": "Transplanted tissue",
+            "example": "tissue from transplanted organ; xenograft tissue",
+            "guidelines": "Should be one of: tissue from transplanted organ, xenograft tissue"
+        },
         "state_of_specimen": {
             "description": "State of the specimen at the time of collection.",
             "type": "object",
--- a/json_schema/type/biomaterial/specimen_from_organism.json
+++ b/json_schema/type/biomaterial/specimen_from_organism.json
@@ -84,15 +84,15 @@
             "guidelines": "If a healthy specimen is sampled from a site adjacent to diseased tissue, enter that tissue's disease here. If no diseased tissue is adjacent to the specimen, leave blank."
         },
         "transplant": {
-            "description": "Description of tissue sampled from a transplanted organ.",
+            "description": "Tissue or organ transplanted from a donor to a recipient.",
             "type": "string",
             "enum": [
-                "tissue from transplanted organ",
-                "xenograft tissue"
+                "Allograft",
+                "Xenograft"
             ],
             "user_friendly": "Transplanted tissue",
-            "example": "tissue from transplanted organ; xenograft tissue",
-            "guidelines": "Should be one of: tissue from transplanted organ, xenograft tissue"
+            "example": "Allograft; Xenograft",
+            "guidelines": "If the sampled tissue or organ were transplanted fill with one of: Allograft, Xenograft"
         },
         "state_of_specimen": {
             "description": "State of the specimen at the time of collection.",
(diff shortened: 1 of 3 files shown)


amnonkhen:
Should `versions.json` be modified as well?

json_schema/type/biomaterial/specimen_from_organism.json
```diff
@@ -83,6 +83,17 @@
             "user_friendly": "Adjacent disease(s)",
             "guidelines": "If a healthy specimen is sampled from a site adjacent to diseased tissue, enter that tissue's disease here. If no diseased tissue is adjacent to the specimen, leave blank."
         },
+        "transplant": {
+            "description": "Tissue or organ transplanted from a donor to a recipient.",
+            "type": "string",
+            "enum": [
+                "Allograft",
+                "Xenograft"
+            ],
```
amnonkhen, line 92: Have you considered using an ontology term? sych as http://purl.obolibrary.org/obo/NCIT_C12981

amnonkhen (changes requested):

json_schema/type/biomaterial/specimen_from_organism.json
ESapenaVentura: not a bad idea - Could live under the class that amnon mentions, although that means we may accept terms we don't expect there

idazucchi: the aim of the transplant field is to highlight if a tissue was sampled from a different donor. This has two implications:
1. the tissue has a different genetic background from other specimens from the same donor
2. the tissue can contain cell from both the transplant donor and the recipient. This is particularly important for a xenografts because some reads will need to be mapped to a different reference genome

The two terms Allograft and Xenograft convey enough information on the two points above. 
The other terms in the ontology branch are subtypes of either, or possibly both since the description doesn't clearly state if the graft is done within the same species (Cadaveric Venous Graft, Organ Graft, Free Arterial Graft). They add context on the medical/surgical procedure but don't add information that's relevant for the two points above and in most cases lack information on whether the tranplant is an allograft or xenograft.

For these reasons I'm against ontologising the field

hannes-ucsc: > I'm against ontologising the field because I don't think that it would add a lot of useful information

As far as I understand it, the motivation behind ontologising a field is to eliminate the need for schema (and metadata) updates when terms are added or updated, and to avoid inconsistencies between the schema enum and any ontologies that might already exist now, or that might be added in the future.

> but rather it would make distinguishing between allograft and xenograft less easy

We will ultimately solve this in Azul/Data Browser. Users will be able to search by an ancestor term and that search will return references to any descendant term. It might be a while until this is implemented but we should take the long-term view here unless there is a good reason not to.

idazucchi: It's true that ontologizing a field can help avoid future updates, but in this case I can't imagine what other field we might want to add. Either the transplanted tissue comes from the same species or it comes from a different one. There might be artificial prothesis but those don't have cells that can be sequenced.

>but rather it would make distinguishing between allograft and xenograft less easy

search by an ancestor term won't enable users to distinguish xenograft from allograft because of the ontology's structure: all terms under [graft](https://www.ebi.ac.uk/ols4/ontologies/ncit/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FNCIT_C12981?lang=en) are siblings and are not under either xenograft or allograft. It is necessary to use xenograft or allograft to specify explicitly if the donor and transplant recipient are from the same species

Screenshot 2023-09-11 at 18 21 13

NoopDog: Hi @idazucchi, I feel this and all other fields should be ontologized if possible. We don't need to allow all values from the ontology.  The graph restriction section should let us specify that we only accept Allograft and Xenograft and their children.

NoopDog: Hi @Wkt8 is there extra effort to ontologize the field? Also, for my knowledge, why do you feel we will not have the benefit of interoperability in this case? Thanks! D

hannes-ucsc:
I assume this is superseded by https://github.com/HumanCellAtlas/metadata-schema/pull/1548 so I am closing it. If I'm wrong, please just reopen.

PR #1537, opened 2023-11-08
Added data_access module for managed access data. Fixes #1535

### LINKED ISSUE #1535: add data access details to project

I would like to request an update to the [`project.json`](https://github.com/HumanCellAtlas/metadata-schema/blob/master/json_schema/type/project/project.json) schema.

I would like to add a new field - `dataAccess` - to this schema to allow data contributors to state how the dataset can be accessed. E.g. Open, Managed

example value
```json
{ 
   "dataAccess": { 
      "type": "All managed access",
      "notes": "data available on GSA databases with accession number HRA000nnn"
   }
}
```

* Field name: dataAccess
  * Field description: data access type
  * Field type: object
  * sub fields: type and notes

* Field name: type
  * Field description: data access type
  * Field type: string
  * Required: yes
  * Examples: All fully open;All managed access;A mixture of open and managed;It's complicated
  * CV or enum: enum

* Field name: notes
  * Field description: notes clarifying the data access type
  * Field type: string
  * Required: no
  * Examples: data available on GSA databases with accession number HRA000nnn
  * CV or enum: no

Part of the Managed Access effort for HCA ingest.
See ebi-ait/dcp-ingest-central#967

Additional tasks:
_should be linked to the relevant development tasks for the ingest backend_
- [ ] @idazucchi to make a report of what is currently recorded in the data access notes - do we want this information published or moved to the admin notes?
- [ ] Once the field is added, existing data, which currently resides in the ingest db and accessible via the API in the `dataAccess` field, should be migrated to the new location.
- [ ] ingest ui should be change to read and write this new field in the correct location of the project json.

FYI @gabsie

ESapenaVentura:
Fixes #1535

related to: ebi-ait/dcp-ingest-central#967

### Release notes

For type/project/project schema:
- Added new required field `data_access`

For module/ontology/data_access_ontology schema:
- Created module


ncalvanese1 (approved):
LGTM

hannes-ucsc:
Can you describe what use cases this change is intended to facilitate? I assumed we were planning to use DUOS and this doesn't change doesn't seem to tie into that.

amnonkhen:
> Can you describe what use cases this change is intended to facilitate? I assumed we were planning to use DUOS and this change doesn't seem to tie into that.

Good point @hannes-ucsc 
We can have another field in addition to `type` and `notes` to hold the DUO code. It can be an enum or an ontology term of children of [data use permission
](http://purl.obolibrary.org/obo/DUO_0000001) or [data use modifier
](http://purl.obolibrary.org/obo/DUO_0000017). Currently, the implementation on the ingest side takes into account the `type` property when authorising an API call. I don't object to use the DUO code for that, but I would like to finish system tests on the change as it is, so I would like to get this schema change live in the dev environment, for which I will need to merge it to the dev branch.

hannes-ucsc:
When I said DUOS, I meant https://duos.broadinstitute.org/

@ncalvanese1 would the addition of a field with the DUO ontology as described by @amnonkhen work for you when it comes to setting up snapshot permissions?

amnonkhen:
> When I said DUOS, I meant https://duos.broadinstitute.org/
I need a project to have an indication whether it is open access or managed access, because I use it do allow or prevent access to metadata in ingest via its REST API. 
@hannes-ucsc Are you objecting to the to `type` and `notes` attributes of the `dataAccess` module, or are you suggesting to add something in addition to them?

hannes-ucsc:
I'm not objecting to anything. I am asking questions.

gabsie:
Hi, @hannes-ucsc @amnonkhen @ESapenaVentura @ncalvanese1 

Basically the purpose of storing with us and passing this information (managed access or open access) to the downstream components will allow us to differentiate how to handle those datasets.

As Amnon has said, this way we will make sure these projects are treated with the required security with us in ingest and down the line with you. So for now with the pilot we are starting, we can mark that project as 'managed access'.

There is no problem to add an additional field/note around DUO codes, but I don't know if at this stage we will always know these from contributors or can assign them correctly. I think we have to continue our current conversations about managed access implementations to decide on the DUO codes. 

Hope this is okay and a good first step.

hannes-ucsc:
Simply marking a project as containing managed-access data is not sufficient information for determining and enforcing *who* should have access to that data. Assuming that in order to answer *that* question, we ultimately want to integrate with DUOS (not DUO) and SAM, the changes proposed here are not taking us in that direction. They will likely need to be backed out and replaced with something else.

For the pilot, why can't we just communicate the set of MA-projects to @ncalvanese1 when they are ready to be imported, instead of burdening the schema with a temporary solution? The set of people who should have access will also have to be communicated to @ncalvanese1 so giving him a few project UUIDs in addition to that set of user identities seems no big deal.

Because I can't attend the biweekly Tuesday meetings, I am likely out of the loop with respect to the pilot. If someone here is the mastermind for the pilot effort, or knows who that person is, please let me know.

amnonkhen:
@hannes-ucsc each component in the DCP will figure out the access rights of its own users. ingest will get information from the DAC in either push or pull mode about users who need access to the project, and protect the api calls accordingly. The same would apply for other components. The access list can change after the project has been exported by ingest, so it is not ingest's responsibility to communicate the access rights down stream. The only thing ingest would communicate is whether the project is a managed access project. Downstream components would query the DAC (or consult a local copy) for managed access projects.

In addition, we need the change in the pilot because ingest uses the schema when it exports files. This is a small change, which we can refine at a later stage.

amnonkhen (approved):

hannes-ucsc:
> each component in the DCP will figure out the access rights of its own users

That's a bit simplistic, I'm afraid, and I must point out, *your* opinion, not some already accepted decision of the DCP/2. 

In my opinion, each component shouldn't have to "figure out access rights", it should *enforce* access, and do so consistently with other components across the platform, and we've already adopted the [specification](https://github.com/HumanCellAtlas/dcp2/blob/main/docs/dcp2_system_design.rst#id74) how this would work in TDR and downstream. You [approved the PR](https://github.com/HumanCellAtlas/dcp2/pull/34) that added that part of the specification. Each component can only *enforce* access if it has the information to do so. To that extent, Azul and Data Browser implement the specification, and tie into TDR for determining who has access. What's missing is the mechanism by which *TDR* determines access and I thought consensus was that we would use DUOS. We need to figure out how this would work, ideally with an addition to the DCP/2 specification.

> This is a small change, which we can refine at a later stage.

If you need a way to temporarily mark a project as managed-access for the pilot, you could just adopt a naming convention for the project description or its short name. That way we wouldn't need to burden the schema with a change that we already know is insufficient for a complete implementation of managed-access for HCA.

hannes-ucsc:
That works for me!

gabsie:
I will tell them to add you to that meeting. Meanwhile, just us need to catch up, so I suggest next week Wednesday 4 or 5pm UK time? Let me know if this works, otherwise please suggest another time.

amnonkhen:
@hannes-ucsc I see 2 scenarios:
1) it is ingest's responsibility to update DUOS with the managed access information when projects are registered in ingest as "managed access"? 

2) managed access registration is recorded on DUOS (not sure yet by whom). HCA components, query DUOS given a request to access a project resource, so that they can enforce access to that resource

I have been under the impression that scenario 2 is the effective scenario. Do you see it differently?

Can you please clarify what you mean?

hannes-ucsc:
> Can you please clarify what you mean?

Specifically, which of my statements is unclear to you?

gabsie:
Hi @amnonkhen and @hannes-ucsc, but also @ncalvanese1   - let's discuss this in a call. I have proposed this Wednesday, **29 Dec**, **4 pm UK time**. Does that work for you?

gabsie:
Hey Hannes, no need for now for the separate meeting. 
Let's as a start get you to join the next HCA Managed access call, that will potentially clarify some aspects of specifically the role of the DACO and how they handle registrations and requests. 
If we still need another meeting after, we can schedule that. 

Gabby

idazucchi:
The fields required for the PR changed slightly due to conversations in the managed access biweekly meeting, so I've updated the body of the PR and added a comment in the original ticket (#1535 ) to reflect what we discussed.


**One note on the ontology module:**
The DUO codes are part of the HCA ontology so in principle we could validate them the way we usually do with the biovalidator keyword graph_restriction. The problem is that the biovalidator doesn't support validation against selected terms but only selected subgraphs; as we discussed with the HCA executive office the only type of access we want to support are those listed in the PR, while using the graph_restriction keyword would allow users to select other terms that are under [general research use](https://www.ebi.ac.uk/ols4/ontologies/duo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FDUO_0000042?lang=en), for example [disease specific research](https://www.ebi.ac.uk/ols4/ontologies/duo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FDUO_0000007?lang=en) which is too narrow for HCA's purpose


For this reason I listed the allowed DUO codes as an enum field, while keeping the usual structure of the ontology module

json_schema/module/ontology/data_access_ontology.json
```diff
@@ -0,0 +1,48 @@
...
+                 "enum": [
+                     "DUO:0000004",
+                     "DUO:0000046",
+                     "DUO:0000042"
+                 ],
+            "user_friendly": "Data access ontology ID",
+            "guidelines": "Must be one of: DUO:0000004, DUO:0000046, DUO:0000042",
+            "example": "DUO:0000004"
+        },
+        "ontology_label": {
```
ESapenaVentura, line 41: May be worth to add an enum here as well - Corresponding to the strings we want to allow with the IDs above

json_schema/property_migrations.json
```diff
@@ -1,5 +1,13 @@
 {
     "migrations": [
+        {
+            "source_schema": "project",
+            "property": "data_access",
+            "effective_from": "18.0.0",
+            "reason": "Data access is now recorded as a mandatory field",
+            "type": "new required property"
+
```
ESapenaVentura, line 9:
```suggestion
```

json_schema/type/project/project.json
```diff
@@ -164,6 +165,12 @@
             "type": "integer",
             "example": "10000; 2100000",
             "user_friendly": "Estimated cell count"
+        },
+        "data_access": {
+            "description": "Information about the data access type for the project.",
+            "type": "object",
+            "$ref": "module/ontology/data_access_ontology.json",
```
ESapenaVentura, line 172: It's only going to be 1 type of data access, won't it?

If we're expecting that, please mark this conversation as resolved :)

ESapenaVentura (commented):
Commented @idazucchi

json_schema/type/project/project.json
idazucchi: Yes, we're not prepared to have different part of the project under managed access and others on open access, so we must use one code per project

json_schema/type/protocol/analysis/analysis_protocol.json
```diff
@@ -58,6 +61,18 @@
...
+            "description": "Name of alignment software used to map FASTQ files to reference genome.",
+            "type": "string",
+            "user_friendly": "Alignment software",
+            "example": "Cellranger; kallisto bustools; GSNAP; STAR"
+        }, 
+        "alignment_software_version": {
+            "description": "Version of alignment software used to map FASTQ files to reference genome.",
+            "type": "string",
+            "user_friendly": "Alignment software version",
+            "example": "v2.0.1; 2.4.2a; v0.45.2"
```
hannes-ucsc, line 75: These changes shouldn't be in this PR.

hannes-ucsc (changes requested):

NoopDog:
This field is meant to be the official statement of how the data is consented. 

It is my understanding that it is a business / legal risk requirement to enforce that this field is 1) required and 2) Explicitly set to NRES to indicate that the data content agreement has no restrictions. 

While we may represent data consents at a file or donor level in the future. This is likely a long way off, and a lot of work so I don't see the value in making this field an array at this point.

To fit the current requirements as I see them, this field needs to be:

1. Required
2. Support all possible data consents, including NRES.
3. Be single-valued.

This sounds like a great discussion topic for the next managed access call.

Cheers,
D

hannes-ucsc:
> The requirement is to represent only a small number of coarse grained consents at the project level including NERS.

If that's true, which I doubt, then it is a misguided and short-sighted requirement in my opinion. It doesn't hurt to look a little bit further into the future.

> For the foreseeable future all data in a project will be required to have the same consent. If multiple consents are required, the project will be split into multiple projects each with a unique consent.

Also a bad idea. This would cause all of the other project-level metadata to have to be duplicated, and break up things that otherwise belong together. It is far easier to allow multiple values in a single field. What I am asking is very easy to do.

> I don't see how this field affects Azul.

It doesn't and no one claims it does.

json_schema/module/ontology/data_access_ontology.json
idazucchi: I don't mind adding an enum for the labels as well but I would like to avoid a case where the DUO id and label are mismatched. I will look into a better way to enforce this constraint

json_schema/type/protocol/analysis/analysis_protocol.json
ESapenaVentura: Apologies, that's my bad - I created the PR against `develop`, which we are not using anymore. Rebased to `staging`

idazucchi:
With the last few commits I addressed two comments which are not visible now:
1. I’ve changed naming from data use to data restriction for all fields/descriptions 
2. Labels: I’ve made the `ontology_label` into an enum and added a set of rules to enforce the correct pairs of labels and ids. I think this is important because otherwise we can get cases where the ID indicates that the dataset is managed access while the label indicates open access and we don’t know which one to trust

In addition to this I’ve made the ontology id required so we can never push a project that lacks the access restrictions. I’ve also made the ontology label required so we have a field that’s human readable, which makes it easier to check that we selected the right DUO code for the project

json_schema/module/ontology/data_use_restriction_ontology.json
```diff
@@ -0,0 +1,94 @@
...
+            "user_friendly": "Data use conditions ontology ID",
+            "guidelines": "Must be one of: DUO:0000004, DUO:0000046, DUO:0000042",
+            "example": "DUO:0000004"
+        },
+        "ontology_label": {
+            "description": "The preferred label for the ontology term referred to in the ontology field. This may differ from the user-supplied value in the text field.",
+            "type": "string",
+            "enum": [
+                "no restriction",
+                "non-commercial use only",
```
hannes-ucsc, line 47: IIRC, there was some talk during our meeting that NCU was to be used as a modifier in combination with GRU, at least in the form to be filled out by contributors. This PR models NCU as a stand-alone enum item. Do we need to change the form or the schema or do are you OK with that inconsistency?

hannes-ucsc (commented):
There was some talk that

json_schema/module/ontology/data_use_restriction_ontology.json
```diff
@@ -0,0 +1,94 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
```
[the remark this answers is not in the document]
ESapenaVentura: It's a bit confusing but yes, for draft 7 the meta-schema `$id` contains a hash at the end https://json-schema.org/draft-07/schema 

Following drafts do not continue with this convention, not entirely sure why

idazucchi:
I restructured the enum field to combine the general research use code with the non-commercial use only modifier. Considering all the rules we’ve added this module won’t behave like other ontology modules I’ve moved to the project modules instead.

json_schema/module/project/data_use_restriction.json
```diff
@@ -0,0 +1,89 @@
...
+            "pattern": "^[0-9]{1,}.[0-9]{1,}.[0-9]{1,}$",
+            "example": "4.6.1"
+        },
+        "DUO_id": {
+            "description": "Type of data use restrictions associated with the project.",
+            "type": "string",
+            "enum": [
+                "DUO:0000004",
+                "DUO:0000042",
+                "DUO:0000042;DUO:0000046"
```
hannes-ucsc, line 30: The use of a custom separator strikes me as hacky since it requires custom parsing on the consumer end. Multiplicity in JSON is natively handled as—I hesitate to bring this up again—arrays. You still restrict the valid term combinations with `allOf` and require `DUO:0000004`.

Alternatively, we could ditch the combination from the contributor form. I personally find it confusing and it makes for a simpler "radio button" form UI instead of the more complicated "only some checkbox combinations are valid" approach.

NoopDog:
Hi folks, 

I wanted to take a second to point out some prior art.

It looks like Terra and the DUOS API encode the meaning of the consents rather than consistently using the ontology terms and ontology terms IDs.

### DUOS API

Looking at the output of the DUOS API, you can see modeling like:

```
 "dataUse": {
      "generalUse": true,
      "hmbResearch": false,
      "diseaseRestrictions": [],
      "populationOriginsAncestry": false,
      "commercialUse": true,
      "ethicsApprovalRequired": false,
      "collaboratorRequired": false,
      "geographicalRestrictions": "",
      "geneticStudiesOnly": false,
      "publicationResults": false
    },
```

Here is an example where ontology term ids _are_ used (for disease):

```
"dataUse": {
      "generalUse": false,
      "hmbResearch": false,
      "diseaseRestrictions": [
        "http://purl.obolibrary.org/obo/DOID_1287"
      ],
      "populationOriginsAncestry": false,
      "commercialUse": true,
      "ethicsApprovalRequired": false,
      "collaboratorRequired": false,
      "geographicalRestrictions": "",
      "geneticStudiesOnly": false,
      "publicationResults": false
    },
```

The DUOS API also gives a text summary for the restrictions like:

```
"translatedDataUse": "Samples are restricted for use under the following conditions:\nData use is limited for studying: brain cancer [DS]\nCommercial use is not prohibited.\nData use for methods development research irrespective of the specified data use limitations is not prohibited.\nRestrictions for use as a control set for diseases other than those defined were not specified."

```
### Terra

In Terra, you can see what looks to be a UI embodiment of this concept in the workspace dataset attributes like:

image

### AnVIL Explorer

In the AnVIL Explorer and Dataset Catalog, the consents are represented. in a single string like:

image

### AnVIL Dataset Catalog

Here is an example of a consent code with explanatory text from the AnVIL Dataset Catalog

image

### Options
Following the examples above, some options for us are:

1. Use a single string to represent the consent.  e.g., "NRES," "GRU," or "GRU-NPU". This is simple, flexible, and future-proofs us to represent any consent in the future. We would validate that the consent string is in the allowed set.

2. Use the DUOS/Broad type approach to semantically represent the specific constraints we care about with a structure like:

**NRES**
```
{
      "noRestrictions": true,
      "generalResearchUse": false,
      "non-commercial-use-only: false
}
```
**GRU**
```
{
      "noRestrictions": false,
      "generalResearchUse": true,
      "non-commercial-use-only: false
}
```

**GRU-NPU**
```
{
      "noRestrictions": false,
      "generalResearchUse": true,
      "non-commercial-use-only: true
}
```

**Translated Data Use**
Optionally for both approaches above, we could add a "translatedDataUse" Field, with values calculated from the input of

NRES - No restrictions
GRU - General research use
GRU-NPU - General research use by not-for-profit organizations only. 
(text would need to be validated of course)


Given that we have basic requirements for representing consents, it seems like the 1 above (e.g., GRU-NPU)  might be easiest all around.

I hope this helps. I am curious what other folks think.

Cheers,
D

NoopDog:
After our discussion, I wanted to propose that we use an enum with the following allowed values:

NRES (DUO_0000004)
GRU (DUO_0000042)
GRU-NCU (DUO_0000042 - DUO_0000046)

Cheers,
D

NoopDog:
For reference, a link to the Data Use Ontology is here https://www.ebi.ac.uk/ols4/ontologies/duo

hannes-ucsc (commented):
I withdraw my vote. You can go ahead and merge without my approval.

https://humancellatlas.slack.com/archives/C01360XN04S/p1709227804910639?thread_ts=1709053675.573539&cid=C01360XN04S

json_schema/type/project/project.json
```diff
@@ -164,6 +165,18 @@
...
+        "data_use_restriction": {
+            "description": "Data use restrictions that apply to the project.",
+            "type": "string",
+            "enum": [
+                "NRES",
+                "GRU",
+                "GRU-NCU"
+            ],
+            "user_friendly": "Data use restriction",
+            "guidelines": "Must be one of: NRES, GRU, GRU-NCU. The use restriction codes are based on the DUO ontology where NRES corresponds to DUO:0000004, GRU corresponds to DUO:0000042, GRU-NCU corresponds to DUO:0000042;DUO:0000046",
```
ESapenaVentura, line 178:
```suggestion
            "guidelines": "Must be one of: NRES, GRU, GRU-NCU. The use restriction codes are based on the DUO ontology where NRES corresponds to DUO:0000004, GRU corresponds to DUO:0000042, GRU-NCU corresponds to a combination of DUO:0000042 and DUO:0000046",
```

src/schema_linter.py
```diff
@@ -16,7 +16,7 @@
 
 required_schema_fields = ['$schema', 'description', 'additionalProperties', 'title', 'name', 'type', 'properties']
 
-allowed_schema_fields = ['$schema', 'description', 'additionalProperties', 'required', 'title', 'name', 'type', 'properties', 'definitions', 'dependencies']
+allowed_schema_fields = ['$schema', 'description', 'additionalProperties', 'required', 'title', 'name', 'type', 'properties', 'definitions', 'dependencies','allOf']
```
ESapenaVentura, line 19:
```suggestion
allowed_schema_fields = ['$schema', 'description', 'additionalProperties', 'required', 'title', 'name', 'type', 'properties', 'definitions', 'dependencies']
```
Not using it anymore :)

ncalvanese1 (approved):

amnonkhen (approved):
KISS! Love it!

PR #1548, opened 2024-02-16
Added transplant_organ field in specimen_from_organism. Fixes #1547

### LINKED ISSUE #1547: Add transplant_organ field in specimen_from_organism

I would like to add a new field - `transplant_organ` - to this schema to allow data contributors to record whether the specimen_from_organism was collected after an organ donation.

This update constitutes a minor change to the schema it affects.

* Field name: transplant_organ
* Field description: Was the specimen collected after extraction for organ transplantation?
* Field type: boolean
* Required: no
* Examples: yes; no
* CV or enum: no

We would like to record if donor organism donated their organs but their organs were considered not fit for transplantation. The quality of the samples is different compared to surgical donors (i.e. biopsy, blood draw) or post-mortem donors.
Is one of the upcoming Tier 1 metadata proposed from HCA Integration Teams & Bionetworks
> The source of the sample (whether the sample comes from a surgical donor; an organ donor; or deceased subject) can result in different cellular profiles and hence batch effects.

arschat:
### Release notes

For `type/biomaterial/specimen_from_organism` schema:
- added `transplant_organ` field
 

### Reviews requested

- This is a minor schema change, all DCP2 [designated reviewers](https://github.com/HumanCellAtlas/dcp2/blob/main/docs/dcp2_operating_procedures.rst#163designated-reviewers) need to review.

### DIFF
--- a/json_schema/type/biomaterial/specimen_from_organism.json
+++ b/json_schema/type/biomaterial/specimen_from_organism.json
@@ -115,6 +115,13 @@
             "type": "object",
             "$ref": "module/process/purchased_reagents.json",
             "user_friendly": "Purchased specimen"
+        },
+        "organ_donor": {
+            "description": "Whether biomaterial was collected after extraction for organ transplantation.",
+            "type": "boolean",
+            "user_friendly": "Organ donor",
+            "guidelines": "Should be one of: yes, or no.",
+            "example": "yes; no"
         }
     }
 }
--- a/json_schema/type/biomaterial/specimen_from_organism.json
+++ b/json_schema/type/biomaterial/specimen_from_organism.json
@@ -117,7 +117,7 @@
             "user_friendly": "Purchased specimen"
         },
         "organ_donor": {
-            "description": "Whether biomaterial was collected after extraction for organ transplantation.",
+            "description": "Was the specimen collected after extraction for organ transplantation?",
             "type": "boolean",
             "user_friendly": "Organ donor",
             "guidelines": "Should be one of: yes, or no.",
--- a/json_schema/type/biomaterial/specimen_from_organism.json
+++ b/json_schema/type/biomaterial/specimen_from_organism.json
@@ -65,6 +65,13 @@
             "guidelines": "The term organ part is very broadly defined here.",
             "user_friendly": "Organ part"
         },
+        "organ_donor": {
+            "description": "Was the specimen collected after extraction for organ transplantation?",
+            "type": "boolean",
+            "user_friendly": "Organ donor",
+            "guidelines": "Should be one of: yes, or no.",
+            "example": "yes; no"
+        },
         "diseases": {
             "description": "Short description of known disease(s) of the specimen.",
             "type": "array",
@@ -115,13 +122,6 @@
             "type": "object",
             "$ref": "module/process/purchased_reagents.json",
             "user_friendly": "Purchased specimen"
-        },
-        "organ_donor": {
-            "description": "Was the specimen collected after extraction for organ transplantation?",
-            "type": "boolean",
-            "user_friendly": "Organ donor",
-            "guidelines": "Should be one of: yes, or no.",
-            "example": "yes; no"
         }
     }
 }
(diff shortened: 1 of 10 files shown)


hannes-ucsc (approved):

NoopDog (approved):

ESapenaVentura (changes requested):
I think this field would be better suited in the Specimen, given multiple specimens can be taken from a donor and not all of them may come from a transplanted organ

json_schema/type/biomaterial/specimen_from_organism.json
```diff
@@ -115,6 +115,13 @@
             "type": "object",
             "$ref": "module/process/purchased_reagents.json",
             "user_friendly": "Purchased specimen"
+        },
+        "organ_donor": {
```
hannes-ucsc, line 119:
```suggestion
        "transplanted_organ": {
```

And unless you have a policy of always appending fields to the schema, should this be inserted after organ_parts?

json_schema/type/biomaterial/specimen_from_organism.json
```diff
@@ -115,6 +115,13 @@
...
+            "description": "Was the specimen collected after extraction for organ transplantation?",
```
hannes-ucsc, line 120:
```suggestion
            "description": "Was the specimen collected after extraction for organ transplantation?",
```

Does it matter if collection happened before or after extraction?

```suggestion
            "description": "Was the specimen collected from an organ about to be transplanted to another organism?",
```

hannes-ucsc:
Note that these are questions/suggestions, not requested changes.

json_schema/type/biomaterial/specimen_from_organism.json
arschat: Hi Hannes, bionetworks are interested in recording this since despite that this is not specimen from a living organism (in most transplant cases) the tissue is kept alive and therefore the transcriptome quality of the sample is in much better condition that to a post-mortem sample.

Therefore, yes, it is important to record that when collection happened after extraction.

json_schema/type/biomaterial/specimen_from_organism.json
arschat: `transplanted_organ` might be interpreted as the specimen was collected from a transplanted organ (after successful transplantation) rather than an organ that was intended for transplantation (transplantation not done).

Thanks for the suggestion about the order. I will insert the field after the organ_part.

hannes-ucsc: I see what you mean. I worry, though, that `organ_donor` will be misinterpreted to mean that the organism the specimen was taken from is an organ donor in general. There is also a potential "collision" of terms in that the donor schema is called donor_organism. I would stay away from "donor" altogether and use a term derived from "transplant". The description field is helpful but is not always available when users look at the metadata downstream, like in a JSON or TSV manifest generated by Azul.

Since this ultimately represents information about a process, i.e., *how* the specimen was collected, should this maybe be a field of the [collection_protocol](https://github.com/HumanCellAtlas/metadata-schema/blob/master/json_schema/type/protocol/biomaterial_collection/collection_protocol.json) entity?

arschat: I see your point, about the word *donor*. In that perspective I would suggest `transplant_organ`. 

About moving to collection_protocol, this field does not describe how the specimen was collected (this can be described by [this](http://www.ebi.ac.uk/efo/EFO_0009124) ontology, in `collection_protocol.method.text`).
This is about how the specimen was treated after the extraction because it was proposed for transplant. Similar fields that describe the process after the collection that affects the quality of the sample exist in the specimen level, like `specimen_from_organism.state_of_specimen.postmortem_interval`.

hannes-ucsc: I like `transplant_organ`.

hannes-ucsc (approved):

amnonkhen (approved):

NoopDog (approved):

ESapenaVentura (approved):
LGTM; Let me know if you need assistance solving the conflicts

PR #1551, opened 2024-03-08
adds DUOS id field

### LINKED ISSUE #1550: Track DUOS project id

`type/project/project.json`

I would like to add a new field - `DUOS_id` to this schema to:
- have a permanent record of a managed access project's DUOS counterpart
- allow the import team to link DUOS' access management system to the project snapshot 
- allow data consumers to find and request access to the project on DUOS

This update constitutes a major change to the schema it affects.

* Field name: DUOS_id
* Field description: A DUOS study accession.
* Field type: string
* Required: yes if `data_use_restriction: "GRU"` or `data_use_restriction: "GRU-NCU"`
* Examples: DUOS-000150, DUOS-000108, DUOS-000114, DUOS-000111
* CV or enum: no
* Guidelines: Managed access projects are registered in DUOS to regulate access. If the project is managed access record the corresponding DUOS id here.

This change will make sure we can trace the DUOS id for each managed access project and can easily link back to DUOS to handle access requests

idazucchi:
Fixes #1550 

#### Rationale

1.  The DUOS id should be required only if the project is managed access (`data_use_restriction = GRU` or `GRU-NCU`)
2. If the project is open access (`data_use_restriction = NRES` ) `DUOS_id` should be forbidden. This way if a project is erroneously marked as open access we can’t export it with a DUOS id.

#### Implementation:
1. The if block allows to conditionally require a field
2. To make sure that `DUOS_id` is disallowed for open access projects it should be conditionally defined for `data_use_restriction = GRU` or `GRU-NCU`. 
This is not possible because `additionalProperties` is set to false, which means that any property not defined in the main property section will fail validation. Removing `additionalProperties : false` is not an option because then any undeclared property would also be valid.
In draft 9 by the keyword `unevaluatedProperties` allows to define properties in subschemas, but we are using draft7 so this option is not available to us right now.
The compromise I made is to require the length of the `DUOS_id` to be 0, so that if the `data_use_restriction` isn’t set correctly the schema will not validate.

### Release notes

For type/project/project.json schema:
- Adds `DUOS_id` field

### DIFF
--- a/json_schema/type/project/project.json
+++ b/json_schema/type/project/project.json
@@ -177,6 +177,30 @@
             "user_friendly": "Data use restriction",
             "guidelines": "Must be one of: NRES, GRU, GRU-NCU. The use restriction codes are based on the DUO ontology where NRES corresponds to DUO:0000004, GRU corresponds to DUO:0000042, GRU-NCU corresponds to a combination of DUO:0000042 and DUO:0000046",
             "example": "GRU"
+        },
+        "DUOS_id": {
+            "description": "A DUOS study accession.",
+            "type": "array",
+            "items": {
+                "type": "string",
+                "pattern": "^DUOS-\\d{1,6}$"
+            },
+            "example": "DUOS-000108; DUOS-000114",
+            "user_friendly": "DUOS id",
+            "guidelines": "Managed access projects are registered in DUOS to regulate access. If the project is managed access record the corresponding DUOS id here."
         }
+    },
+    "if" : {
+        "properties": {
+            "data_use_restriction": {
+                "enum": [
+                    "GRU",
+                    "GRU-NCU"
+                ]
+            }
+        }
+    },
+    "then" : {
+        "required": ["DUOS_id"]
     }
 }
--- a/json_schema/type/project/project.json
+++ b/json_schema/type/project/project.json
@@ -178,7 +178,7 @@
             "guidelines": "Must be one of: NRES, GRU, GRU-NCU. The use restriction codes are based on the DUO ontology where NRES corresponds to DUO:0000004, GRU corresponds to DUO:0000042, GRU-NCU corresponds to a combination of DUO:0000042 and DUO:0000046",
             "example": "GRU"
         },
-        "DUOS_id": {
+        "duos_id": {
             "description": "A DUOS dataset id.",
             "type": "string",
             "pattern": "^DUOS-\\d{1,6}$",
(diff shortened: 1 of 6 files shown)


ncalvanese1 (approved):

json_schema/type/project/project.json
```diff
@@ -177,6 +177,37 @@
...
             "guidelines": "Must be one of: NRES, GRU, GRU-NCU. The use restriction codes are based on the DUO ontology where NRES corresponds to DUO:0000004, GRU corresponds to DUO:0000042, GRU-NCU corresponds to a combination of DUO:0000042 and DUO:0000046",
             "example": "GRU"
+        },
+        "duos_id": {
+            "description": "A DUOS dataset id.",
+            "type": "string",
+            "pattern": "^DUOS-\\d{1,6}$",
+            "example": "DUOS-000108; DUOS-000114",
+            "user_friendly": "DUOS id",
+            "guidelines": "Managed access projects are registered in DUOS to regulate access. If the project is managed access record the corresponding DUOS id here."
```
hannes-ucsc, line 187:
```suggestion
            "guidelines": "Managed-access projects are registered in DUOS to regulate access. If the project is managed-access, record the corresponding DUOS ID here."
```

hannes-ucsc (approved):

NoopDog (approved):
Nice! 🚀

json_schema/type/project/project.json
```diff
@@ -177,6 +177,37 @@
             "user_friendly": "Data use restriction",
             "guidelines": "Must be one of: NRES, GRU, GRU-NCU. The use restriction codes are based on the DUO ontology where NRES corresponds to DUO:0000004, GRU corresponds to DUO:0000042, GRU-NCU corresponds to a combination of DUO:0000042 and DUO:0000046",
             "example": "GRU"
+        },
+        "duos_id": {
+            "description": "A DUOS dataset id.",
+            "type": "string",
+            "pattern": "^DUOS-\\d{1,6}$",
```
amnonkhen, line 184: Is it between 1 & 6 digits or a 6 digit number?

amnonkhen (approved):

json_schema/type/project/project.json
idazucchi: It's a 6 digit number

idazucchi: I've updated the regex, thanks for pointing this out!

PR #1566, opened 2024-08-08
Added smoking related fields in medical_history module. Fix #1565

### LINKED ISSUE #1565: Add smoking related fields in medical_history module

 I would like to request an update to the `medical_history.json` schema.

I would like to add three new fields - `smoking_status`/ `smoking_pack_years`/  `years_since_smoking_cessation` - to this schema to allow data contributors to collect precise and measurable information about donor's smoking history. 
Existing field `smoking_history` should be removed since there is overlap with `smoking_pack_years`.

This update constitutes a **major** change to the schema(s) it affects.

1. **smoking_status**
    * Field name: smoking_status
    * Field description: Whether the individual is actively, was formerly or never consumed smoking tobacco products like cigarettes, cigars, pipe, betel nut chewing etc.
    * Field type: string
    * Required: no
    * enum: active; former; never
    * Examples: active; former; never

1. **smoking_pack_years**
    * Field name: smoking_pack_years
    * Field description: Estimated number of packs (20 cigarettes) smoked per day multiplied by the number of years the individual was smoking.
    * Field type: number
    * Required: no
    * Examples: 4.55; 12; 49.5
    * CV or enum: no

1. **years_since_smoking_cessation**
    * Field name: years_since_smoking_cessation
    * Field description: If smoking status is "former", specify the number of years since smoking cessation.
    * Field type: integer
    * Required: no
    * Examples: 1; 4; 12
    * CV or enum: no

Many bionetworks (currently 5) are interested in recording the smoking history of the donor in a standardised way as part of their Tier 2 metadata. 

Although schema already has the `medical_history.smoking_history` field that records _"Estimated number of cigarettes smoked per day."_, this is a free text field that we can't standardize and use as a measurable way, and the way of measurement suggested does not take into account the number of years the individual smoked.

Lung bionetwork has proposed this 3 field way of recording this, and we expect these (or at least the status) to be requested for Tier 2 by most of the bionetworks. Since `smoking_pack_years` is very similar to the existing `smoking_history` we also suggest this field to be removed.

arschat:
### Release notes

For `medical_history.json` schema:
- `smoking_status`
- `smoking_pack_years`
- `years_since_smoking_cessation`
- removed `smoking_history`
 
### Reviews requested

- Need 5 Reviewers to approve because this is a major update

### DIFF
--- a/docs/jsonBrowser/module.md
+++ b/docs/jsonBrowser/module.md
@@ -473,7 +473,9 @@ Property name | Description | Type | Required? | Object reference? | User friend
 --- | --- | --- | --- | --- | --- | --- | --- 
 alcohol_history | Estimated amount of alcohol consumed per day. | string | no |  | Alcohol history |  | 3-6 alcohol units/day; 1 drink per day
 medication | Medications the individual was taking at time of biomaterial collection. | string | no |  | Medications |  | Naproxen 500mg/day; Citalopram 20mg/day
-smoking_history | Estimated number of cigarettes smoked per day. | string | no |  | Smoking history |  | 20 cigarettes/day for 25 years, stopped 2000
+smoking_status | Whether the individual is actively, was formerly or never consumed smoking tobacco products like cigarettes, cigars, pipe, betel nut chewing etc. | string | no |  | Smoking status | active, former, never | Should be one of: active, former, never.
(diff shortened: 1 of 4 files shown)


arschat:
Dropping PR due to wrong branch naming.

Issue #1583, opened 2024-10-09
Extend atlas names in the hca_bionetworks


arschat:
**For which schema is a change/update being suggested?**

I would like to request an update to the module `hca_bionetwork.json` schema.

**What should the change/update be?**

I would like to extend the enum for atlas names in `hca_tissue_atlas`.

This update constitutes a patch change to the schema it affects.

**What field(s) need to be changed?**

* Field name: `hca_tissue_atlas`
* Previous CV or enum: Blood; Retina; Lung; Kidney; Gut; Eye; Brain
* Additional CV or enum: Adipose; Breast; Development; Eye - Other; GDN; Gut; Heart; Immune; Kidney; Liver; Lung; MSK; ORCF; Organoid-Endoderm; Organoid-Neural; Pancreas; Retina; Reproduction; Skin
* Overall: Adipose; Blood; Brain; Breast; Development; Eye; Eye - Other; GDN; Gut; Heart; Immune; Kidney; Liver; Lung; MSK; ORCF; Organoid-Endoderm; Organoid-Neural; Pancreas; Retina; Reproduction; Skin

**Why is the change requested?**

Since this information can be shown in portal for each project, we need to update this metadata for all the projects we have this info.

PR #1584, opened 2024-10-09
Extend atlas names

arschat:
### Release notes

For `hca_bionetwork.json` schema:
- Extend the controlled vocabulary of atlas names in `hca_tissue_atlas`
#1583 

### Reviews requested

- Need 5 Reviewers to approve because this is a patch update


json_schema/update_log.csv
```diff
@@ -1 +1,2 @@
 Schema,Change type,Change message,Version,Date
+module/project/hca_bionetwork,patch,Extend atlas names in hca_bionetwork,,
```
ESapenaVentura, line 2: We should revisit what minor/patch versions are like. Thanks for following the guide https://github.com/HumanCellAtlas/metadata-schema/blob/master/docs/evolution.md#schema-versioning but I think extending an enumeration should be considered a minor patch, as in my opinion any change in the schema that involves change in the validation rules should be at least a minor change. Anyways, this is good to go as per our specs! 🥇

ESapenaVentura (approved):

json_schema/module/project/hca_bionetwork.json
```diff
@@ -48,17 +48,32 @@
...
-                "Lung",
-                "Kidney",
-                "Gut",
+                "Brain",
+                "Breast",
+                "Development",
                 "Eye",
-                "Brain"
-            ],
+                "Eye - Other",
```
NoopDog, line 57: You can remove "Eye - Other"; this will be refined as the atlas becomes more focused.

NoopDog (changes requested):
We can remove Eye-Other for now

json_schema/module/project/hca_bionetwork.json
arschat: Thank you, Dave! Done.

amnonkhen (approved):

hannes-ucsc (approved):

NoopDog (approved):

PR #1597, opened 2024-12-12
Fix schema_linter.py TypeError and OLS api

### LINKED ISSUE #1596: Bug report: src/schema_linter.py failing
Since #1575 PR in [json_schema/system/file_descriptor.json](json_schema/system/file_descriptor.json) where a new property `dri_uri` was introduced, we get an error in schema linter script triggered by github actions

> Run python ./src/schema_linter.py
Traceback (most recent call last):
Linting [8](https://github.com/HumanCellAtlas/metadata-schema/actions/runs/12273465137/job/34244292014#step:5:9)9 schemas
  File "/home/runner/work/metadata-schema/metadata-schema/./src/schema_linter.py", line 355, in <module>
    return_msg = linter.lintSchema(s, ols_api)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/metadata-schema/metadata-schema/./src/schema_linter.py", line 168, in lintSchema
    errors.append(schema_filename + ".json: Type `" + properties[property]['type'] + "` is not a valid JSON type.")
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
TypeError: can only concatenate str (not "list") to str
Error: Process completed with exit code 1.

The linter is failing because `properties[property]['type']` which is expected to be a `str` is a `list` and cannot be concatenated to print the error.

arschat:
Fix #1596 

## Fix TypeError
[File descriptor schema](https://github.com/HumanCellAtlas/metadata-schema/blob/master/json_schema/system/file_descriptor.json) has a `list` in properties/drs_uri/type while schema_linter.py expects a `str`. 

In this PR we will treat all types as list and allow value 'null' as well

## OLS API update
As of November 2024, we are now use the [OLS4 API](https://www.ebi.ac.uk/ols4/) which includes previously used HCAO. 
In this PR the OLS API url is replaced as well.

## Human readable json script
During review of this PR, a bug in the [human_readable_json.py](https://github.com/HumanCellAtlas/metadata-schema/blob/master/src/human_readable_json.py) script was discovered and fixed.
- We might have numeric values as enum values, and this was breaking a `join` function.


ESapenaVentura (approved):
Looks good to me! 

Also great job catching the issue with the JSON schema human readable script :)

Issue #1600, opened 2025-01-08
Replace enum value in hca_bionetwork name


arschat:
**For which schema is a change/update being suggested?**

I would like to request an update to the module `hca_bionetwork.json` schema.

**What should the change/update be?**

I would like to change one atlas name in `name`.

This update constitutes a major change to the schema it affects.

**What field(s) need to be changed?**

* Field name: `name`
* Previous enum value: `Oral & Craniofacial`
* New enum value: `Oral and Craniofacial`

**Why is the change requested?**

Schema has to be aligned with [Data Portal](https://data.humancellatlas.org/hca-bio-networks/oral) nomenclature and it's better to avoid usage of special character.

PR #1601, opened 2025-01-08
Replace enum value in hca_bionetwork name.

arschat:
#1600 

### Release notes

For `hca_bionetwork.json` schema:
- Remove enum value `Oral & Craniofacial`
- Add enum value `Oral and Craniofacial` 

### Reviews requested

- Need 5 Reviewers to approve because this is a major update

### DIFF
--- a/json_schema/update_log.csv
+++ b/json_schema/update_log.csv
@@ -1,5 +1,2 @@
 Schema,Change type,Change message,Version,Date
-module/project/hca_bionetwork,major,"Removed enum value in hca_bionetwork.name. Fix #1600",,
-module/project/hca_bionetwork,patch,"Added enum value in hca_bionetwork.name. Fix #1600",,
-type/project/project,major,"Removed enum value in hca_bionetwork.name. Fix #1600",,
-type/project/project,patch,"Added enum value in hca_bionetwork.name. Fix #1600",,
+module/project/hca_bionetwork,major,"Changed enum value in hca_bionetwork.name. Fix #1600",,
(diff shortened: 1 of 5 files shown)


json_schema/update_log.csv
```diff
@@ -1 +1,5 @@
 Schema,Change type,Change message,Version,Date
```
ESapenaVentura, line 1: Couple of things!

1. You only need to indicate the Major change (Changed enum value...), no need to indicate the patches here, since you're already moving a major version over!
2. Since the hca_bionetwork module is loaded in project, there's no need to indicate it in the `update_log.csv`! the script automatically identifies and updates schema dependencies :)
3. Also, for the future, you need to order the messages by type (major, minor, patch) in that order (that's due to a constraint within the script). You can read more about it in the [committers.md](https://github.com/HumanCellAtlas/metadata-schema/blob/master/docs/committers.md#specific-how-to-for-making-changes) doc, point 5 on the section I linked!

ESapenaVentura (changes requested):
A couple of changes to the update_log, other than that LGTM

ncalvanese1 (approved):

hannes-ucsc (approved):

NoopDog (approved):

json_schema/update_log.csv
arschat: Thanks a lot for the comment Enrique! Applied in 54d04ef362e555c108f169d5f6bf056df2efed9e.

ESapenaVentura (approved):

